Issue #197, opened 2021-09-26
index.php - Breadcrumbs link to index.html


Brodur:
index.php has a breadcrumb trail that links Dashboard to index.html, which does not exits.

johnnyq:
Thank you for the correction @Brodur I'll go ahead and merge your code soon, much appreciated.

PR #198, opened 2021-09-26
Fix broken link in index.php

Brodur:
Fix for issue #197


Brodur:
potentially you might consider changing the href to dashboard.php instead of index.php - I just changed the file extension from html to php.

PR #228, opened 2022-01-03
Check that config.php is writable at the start of setup.

wrongecho:
_My first pull request, please be gentle._

Whilst setting this up on my Linux VM, I noticed that the setup would just continuously loop. Apache did not have access to write files into my directory (chmod 755, as per general best practice). 

I've just added a small check that we can write config.php.
I did see you've got a setup_checks but this doesn't seem to be in use, from what I could see.

(Also please ignore the test file I made whilst setting up git... I don't know how to remove it from this pull request, but its deleted....)

### DIFF
--- a/add_certificate_modal.php
+++ b/add_certificate_modal.php
@@ -32,12 +32,12 @@
           </div>
           
           <div class="form-group">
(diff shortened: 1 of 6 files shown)


wrongecho:
To relog.

PR #230, opened 2022-01-04
Add support for storing client certificate public key

wrongecho:
This commit allows you to store the public key for a client certificate. 

If you store a key in the x509 format, it will also read and update the expiry date/issuer fields for you as well. As such, I've removed the "required" html tags from these fields.


johnnyq:
Fantastic Work! @wrongecho Thank you for your contribution!! will merge here shortly

PR #231, opened 2022-01-04
Link ticket assets

wrongecho:
Allow linking of an asset to a ticket via the edit modal.
Haven't added it to the ticket creation modal at this stage - feel like it disrupts the flow - but could?


johnnyq:
Excellent work, I meant to getting around to this but forgot all about this thank you for your help @wrongecho

PR #234, opened 2022-01-04
Bug fix: Default alert type

wrongecho:
Set default alert type to bootstrap primary. Sometimes we set alert_message in the users session from post.php, but not alert_type. This can cause issues as we rely on the type to display any alert/notification at all.


johnnyq:
Excellent Catch Thanks @wrongecho

PR #233, opened 2022-01-04
Tidy/Bugfix tickets

wrongecho:
Added checks to ensure variables are defined before they are called.

Improved ticket filtering so the all/open/closed buttons work.
Note: If you search/filter by date, you then have to re-filter for open/closed. Partially resolves #224 but needs a little further work.

Issue #181, opened 2021-03-30
Feature Request: Deleting/archiving records should prompt for confirmation


Malcharias25:
I consider this a bit of a bug, but perhaps it fits better as a feature request... I'll leave that to you to decide.

I would like to see a feature implemented where, when someone clicks the Delete option in any menu (client records, documents, assets, etc.) they get a prompt asking to confirm the action - perhaps with a check box or the like. This will help reduce the chances of accidental deletion of records (which I've had happen a couple of times already as I've been inputting information!)

Additionally, it would be nice if there was a user role that had read-only access to the data - this could be used to give clients logins to their specific records, for instance.

johnnyq:
Yeah I must admit @Malcharias25  I have done the same, ill go ahead and implement this feature

johnnyq:
This is complete closing

PR #236, opened 2022-01-05
Client bugs

wrongecho:
Hey, me again :)

Few quick bug fixes.

- Require a primary tech contact be associated when client is created. This prevents an error when ticekts are created from the "general" management page (as no contact is prompted, and one does not exist, the ticket fails to be created).
- Ensure the canned_date is defined in the clients list (clients.php) page, this just prevents a load of PHP's errors (similar to #233)
- Prompt for confirmation when deleting a client from their "banner/summary" page (the bit with contact/billing/support summaries. Contributes to resolving #181, but we still need confirmations for other actions.

### DIFF
--- a/add_client_modal.php
+++ b/add_client_modal.php
@@ -144,12 +144,12 @@
             <div class="tab-pane fade" id="pills-contact">
 
               <div class="form-group">
-                <label>Primary Contact</label>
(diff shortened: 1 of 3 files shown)


johnnyq:
Good Catch and thank you @wrongecho for your commit!!

Issue #201, opened 2021-11-06
A few ideas/questions


DarkObscurity:
Hey!

I love the product and have been looking around  in it for a little and a few questions/ideas came to mind!

I have been working quite a lot in IT-Glue in the past and for my own purpose I was looking for an open source alternative and this is the only project so far that comes close to it! So I'm really grateful for that!

I did have some questions regarding the documentation part.
IT-Glue is all about documentation and having the tools to do so in a easy to access way>

When I was looking around in the asset documentation I noticed that the way this is documented is in tiny pop-up windows instead of documentation pages with things like related items and I was  curious if this might change in the future or of this is how it's going to stay?

I personally value proper documentation a lot with the ability to create guides on the places where people expect them (For example when opening a specific server asset, it would be the perfect place to have any guides/special needs or manuals located in a full page that opens.) 

I also saw that the location feature is limited to the US, and I understand that it might be a lot to add to have the whole world in there, but maybe it might be an idea to allow people to just type in the information them self for the time being? I don't know if this will break anything though...

And are there any plans for adding custom side-bar layouts with page lay-outs? IT-Glue has a really nice system with a sort of customer lay-outs where for example you could create documentation sections for things like Microsoft 365. These pages then contain certain fields/drop down lists/options to be filled in so one can, for each customer, fill in for example what kind of Microsoft 365 services are being used and the lay-out can be distributed across all customers so a baseline can be set.

Long story short, I am curious if the documentation side will be extended in the future! :)

johnnyq:
Thank you for the kind words @DarkObscurity 

1. Yeah I agree, the tiny modal window for documentation is quite limited. I will eventually make this full screen and allow folder, subfolder categorization of documents along with PDF exportation some time in the very near future.
2. I think that would be great for assets,in fact we may be able to make an asset documentation relationship to the documents section. This will also be done
3. We have recently added full country support see 72cf61ab61c3b91324ea71b9795cf978abdf6ff0
4. I have never actually used ITGlue and have only seen some screenshots so im not fully aware of this feature, but it sounds neat and would love to implement something similar.

Thank you for your ideas and kind words, keep them flowing.

DarkObscurity:
> Thank you for the kind words @DarkObscurity
> 
> 1. Yeah I agree, the tiny modal window for documentation is quite limited. I will eventually make this full screen and allow folder, subfolder categorization of documents along with PDF exportation some time in the very near future.
[...]

Issue #174, opened 2021-03-11
Feature Request - Tree Level Document Support


randysrs:
any roadmap ? need a howto/general KB folder at the top level . i get you have a document folder in each client which is cool , but would like to have a option for a tree level

my thought process is we have 50 +documents from LOB/SOP docs, would like to organize them by folders , rather then search 50 docs for those .
thoughts

example of structure

Client Docs

Employee documents
    LOB documents
    General Howtos

upper level Document / KB area would be above all Clients 

Document/KB Area
   Linux Howtos
   Windows Howtos
   General Howtos

function to create additional folder structure
can someone give me some pointers where i can learn more about write that extra code 

Thanks 
Randy

johnnyq:
@randysrs Good idea, but no real roadmap or docs as of yet. Unfortunately i've been quite busy running my MSP lately and has caused me to step back on development on the ITFlow. Looking forward to getting back into to the flow.

wrongecho:
You can now organise documents into folders.

![image](https://user-images.githubusercontent.com/32306651/208251965-e412aa4c-e920-496a-8674-b7622b3beaac.png)

johnnyq:
Complete closing

PR #239, opened 2022-01-05
Add tag functionality to documents

wrongecho:
Hello!

I was looking at organizing documents better and noticed you're planning on adding folders and relationships with assets (#174 / #201) in the future. I've added functionality here to allow tagging of documents in conjunction with this.

It could probably use a little extra work but you get the core concepts.

Hope this is okay. Please let me know your thoughts.


johnnyq:
Wow Fantastic @wrongecho I just can't keep up with you!! haha Very good work!!

PR #240, opened 2022-01-06
Update ticket status in database after a ticket reply.

wrongecho:
When setting the status to working/on hold, the status does not change as it was not set in the post.php add_ticket_reply.


johnnyq:
Excellent merging now

PR #246, opened 2022-01-07
Update client_tickets.php

aftechro:
- re-arranged the table items display


johnnyq:
Fantastic! looks very good @aftechro Thank you!

PR #247, opened 2022-01-07
Update client_contacts.php

aftechro:
added the info of the contact type - for approval procedures/purposes


johnnyq:
Alright! good idea! @aftechro it gives that little nudge to add a contact if none are there

PR #249, opened 2022-01-07
API Changes - General refactor & Add asset via API

wrongecho:
- Refactored the API key auth to be separate from the API functions. Also added a 401 response code to missing/incorrect key.
- Added functionality to add a new asset to the system via the API - e.g. as the final stages of a build script, auto-register the asset
- Updated readme with API instructions


johnnyq:
Excellent! Nice work and very much needed Thanks @wrongecho

PR #251, opened 2022-01-07
Fix user initials in ticket reply

wrongecho:
Another quick one.
I need to try and bunch these together, sorry!


johnnyq:
Good catch!!

PR #253, opened 2022-01-07
Set PHP session cookie to be httponly

wrongecho:
Set the session cookie to be httponly
https://owasp.org/www-community/HttpOnly


johnnyq:
Very nice I did not know about this, its looked weird at first but I looked it up and its a thing to help prevent XSS Thank you!!

Issue #225, opened 2021-12-28
Feature Request: Ticketing System Improvements


athornfam2:
* Create a way to send tickets from customers through e-mail and then parse base on domain name
* Add more to reporting (Average Response and Resolution Time, Service Requests, Ticket satisfaction)
* Add a "smiley response" (customer feedback based on ticket which will eventually be stored under their customer ID)
* Candid responses
* Round-robin ticket selection
* Waiting for response/customer - responses
* Portal for customers to check on tickets
* Portal for customers to open new tickets (custom entry form)
* Mass change ticket entry (example - closing PRTG alerts)
* Ability to post internal notes - non-customer sending
* SLA plans
* Category selection based on subject or body input (keyword search)
* Daily or weekly reports to helpdesk lead/manager based on tickets that haven't received a response/been updated/stale

johnnyq:
All good requests but have to work on most important things first keep em coming

wrongecho:
If I could add to this just as somewhere structured to track my ideas...
- Scheduled tickets (automatically raised weekly/monthly)
- Time tracking
- Merging
- Linking tickets to a master ticket
- Different ticket views/filters (e.g. my tickets, unassigned tickets, my team's tickets)
- Ticket types: incident/service request

athornfam2:
@wrongecho check out ticket #221

wrongecho:
@athornfam2, thanks for pointing that out!

wrongecho:
Moved to https://forum.itflow.org/d/441-moved-from-github-issues-feature-request-ticketing-system-improvements-225

PR #254, opened 2022-01-08
Scheduled tickets

wrongecho:
Added the start of functionality for scheduled tickets - weekly/monthly currently.
Tickets are created on the client ticket page (is this the best place?), and are created by cron.

Haven't added a UI to manage/remove scheduled tickets yet (Consider this more a proof of concept). Not sure how we want to do this. I don't want the ticket page to get messy. Maybe it's better to hide these options in the settings as they're more of a "set and forget"? Would appreciate your thoughts @johnnyq?
Also, if you could reposition the code in cron where you feel it fits best, please?

( #225)


johnnyq:
This will be very useful and yes I believe the client ticket page would be the best place as is. I'm not sure where the best placement for scheduled tickets listing to be. Maybe a separate menu option for it like recurring invoices, the cron code seems like its in a good place but ill take another peek after I merge the code. Good work!!

Issue #221, opened 2021-12-27
Feature Request: Time Entry


athornfam2:
Would like to request a feature. A popular feature I would use in CW was to track time based on ticket entry at the comment or sent response. When I went to go in to make a ticket and then track the time I didn't see that function unless its hidden in the system settings.

Feature Request: Add a time based entry system at the ticket level.

johnnyq:
good idea! that would make a great additon, ill try to get this added very soon.

aftechro:
That`s a tricky one, as we all know there are times when you can answer right away with a solution, or you may need to do a research in getting and trying different solutions. 

The way i work, probably like many others... i dont count the time spent on replying to a ticket instead i`m using this system:
- for maintenance clients - support is included, no time to record
- for non-maintenance - there will be a minimum charge of 0.30 minutes for any type of work (email, phone, remote) - so even though the call is 5 minutes, or remote task is 10 minutes, i`ll still charge 30 minutes for it. Over 30 minutes, i`ll apply the hour rate and so on...in the end, i`m only charging for the actual task itself

But of course, i`ll always do that with common sense - i wont charge if i`m just replying or sharing some knowledge over the phone or email

What i will like to have on the ticketing system. are few buttons: 
-  **add to calendar** - either itflow calendar or integrate with google - (maybe close/resolve ticket and add to callendar) 
-  **invoice** - to open the new invoice - and the subject to start with ticket number eg: #2343 Printer not working - so the invoice reflect the ticket which can be verified on both sides (MSP and customer)

example as i`ve changed it on the osticketing
![image](https://user-images.githubusercontent.com/38830718/147582129-1d221359-01f2-499c-b11d-a4966e53f113.png)

athornfam2:
@aftechro I think my request was more to just have a time entry in general as each MSP and tech (I used a tech SysAdmin) has different formula's to their work. I wouldn't want any entry's to be fixed (ex .25 .50 .75)... Like you said that phone call only may be 3 to 5 minutes... So I still want a time entry but make it a NB (non-billable charge) but still counts towards tech time and as a response.

The thinking in this is similar to how CW Manage works.

wrongecho:
I too like the idea of the timer that just counts up as you work, but you can edit it.
Similar to this: https://docs.servicenow.com/bundle/rome-platform-administration/page/administer/time/image/TimeTimeworked.png

wrongecho:
[...]

PR #255, opened 2022-01-08
Ticket improvements

wrongecho:
Mainly from #225 / #221 (not 255, oops!).
- Add db structure for customer feedback (I like the idea of Good/OK/Bad smiley faces)
- Add support for time tracking
  - You may want to move the JS into a separate file
- Add support for public/internal updates (I've used the existing "Email update to client", hope that's okay :) 
  - I've adjusted the internal notes card background, not sure if you'll like the colour though.
- Small edit to setup.php

What are your thoughts on moving the ticket reply box above the ticket comments? After you get about 5/6 replies you really have to start to scroll. Would make sense as the latest replies are at the top as well.

### DIFF
--- a/db.sql
+++ b/db.sql
@@ -999,6 +999,7 @@ DROP TABLE IF EXISTS `ticket_replies`;
 CREATE TABLE `ticket_replies` (
   `ticket_reply_id` int(11) NOT NULL AUTO_INCREMENT,
   `ticket_reply` longtext NOT NULL,
+  `ticket_reply_type` varchar(9) NOT NULL,
   `ticket_reply_created_at` datetime NOT NULL,
   `ticket_reply_updated_at` datetime DEFAULT NULL,
   `ticket_reply_archived_at` datetime DEFAULT NULL,
--- a/db.sql
+++ b/db.sql
@@ -1000,6 +1000,7 @@ CREATE TABLE `ticket_replies` (
   `ticket_reply_id` int(11) NOT NULL AUTO_INCREMENT,
   `ticket_reply` longtext NOT NULL,
   `ticket_reply_type` varchar(9) NOT NULL,
+  `ticket_reply_time_worked` time DEFAULT NULL,
   `ticket_reply_created_at` datetime NOT NULL,
   `ticket_reply_updated_at` datetime DEFAULT NULL,
   `ticket_reply_archived_at` datetime DEFAULT NULL,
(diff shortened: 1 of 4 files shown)


johnnyq:
- I do too, its simple
- We have an app.js file for this but thats really global I dont see any problem including it directly into the php file
- That sounds fine, ill take a look at the color after i Merge
- nice wording

I agree I actually had the ticket reply box on top of the comments at one point, but I agree I think moving it back will be ideal especially as more commend are posted or we can move the whole ticket reply to a modal which would make the whole ticket details page less cluttered, how do you feel about reply using modal instead?

Thank you

wrongecho:
Thank you! :)

In terms of the ticket reply, I'd personally prefer just keeping it on the page rather than in a separate popup, just at the top. I think it makes it much easier to work with everything on the same page, can just scroll down a little and see the last update. 
We could probably do with adjusting the default height of the reply box - or showing/hiding it with javascript?

I did have a couple of other questions for you please:
- Documentation: this is currently on the itflow site, is there a way to contribute to / add general documentation?
- Roadmap: Besides the issues open here, do you have a general idea of how you want this to progress - and the structure for that progression (via issues/a kanban, etc)? 
  - I just ask as I'm basically adding/adjusting things I like that are in other software. Unless you're essentially open to adding almost anything that might be useful/helpful on the fly, or would you prefer issues are raised first? I'm fairly new to contributing to open source stuff, so just trying to do so in the most efficient way :)
- License: Have you considered adding a project license like GNU GPLv3? This just makes it "official" that this project is open-source, etc. 
  - https://choosealicense.com/licenses/gpl-3.0/

johnnyq:
I agree with you there, lets keep it the way it is for now I have it above the ticket replies

1. Register yourself as a user on ITflow and ill promote you to admin. to make changes to the CMS
2. No Roadmap really just really trying to make everything existing pretty solid and bug free but new feature are welcome, but still keeping it simple and minimal.
3. GNU GPLv3 is what this project is, Maybe adding it to the readme would be good

Don't sell yourself short you had be convinced you were a pro

Also send me you contact details you can use the contact form on pittpc.com.

Thank again

wrongecho:
Reply box definitely looks better at the top. I like what you've done with the replies card outline too, looks very neat but still easy to see what notes are public/internal.
Hahaha, thank you for the compliment :)
I've made an account and submitted the form at pittpc.com. 

Cheers!

PR #256, opened 2022-01-08
Add note when ticket is closed (to easily show who closed it)

wrongecho:


johnnyq:
What so the tech automatically gets a minute to their time when they close a ticket? haha just kidding Thanks!!

wrongecho:
> What so the tech automatically gets a minute to their time when they close a ticket? 

Someone has to look out for the techs here, hahaha.

PR #258, opened 2022-01-09
Secure Cookies

wrongecho:
Set Secure flag on cookies, so they are only sent in HTTPS requests by default.
This can be overridden by setting $config_https_only = FALSE in the config.php file.

Note: Browsers seem to bypass this check for localhost sites.


johnnyq:
Fantastic Security Fix!!

PR #259, opened 2022-01-09
Misc - small changes to document tagging/files

wrongecho:
Few small changes:
- Delete document tag associations when a document is deleted (before this was leaving orphaned tag associations)
- Import assets CSV now checks for CSV files, and their length
- Asset import CSV and file add modal now suggest accepted file types to the user


johnnyq:
Nice Work!!

PR #263, opened 2022-01-09
Hide reply box if ticket is in a closed state

wrongecho:


johnnyq:
Sometime the simplest things can get missed haha good job

Issue #264, opened 2022-01-10
Asset login encryption thoughts


wrongecho:
Just some thoughts on how ITFlow is handling the encryption of asset/login passwords (not like login to the app itself).

**No salt when encrypting passwords**
If you encrypt two login entries with the same password, you get the same output. Presumably, this is due to no salt/IV. 
We should probably be salting passwords?

**Storing AES key in database**
Currently, we're storing the AES encryption key in the database. Theoretically, if an attacker were to gain access and dump the database, they have the key they need to decrypt all passwords.

I've done a little research and the general consensus seems to be to use a HSM (which probably isn't an option for a lot of us). 
Other than using a HSM, there are a few options. There's a good answer here: https://security.stackexchange.com/questions/12332/where-to-store-a-server-side-encryption-key/12334#12334

One way of countering this may be to encrypt the AES key with a shared passphrase (that you communicate to your IT team) and then prompt for this phrase when you want to decrypt a password. 
This passphrase is potentially then stored in the session/DB/apc(?) so the user isn't constantly prompted for it to see every entry - as long as it is destroyed after the user's session ends. This isn't perfect either (as the key is still plain text at some point - security vs usability here) but substantially reduces the attack surface?

Alternatively, if we don't want to make users have to remember a shared secret, we could encrypt the AES key for each user individually with their account password and decrypt it at login. This is slightly more usable but then requires users to have really good passwords. Still the question of whether to store the shared passphrase or prompt for their password every time to decrypt it.

I think at the very least we should move the AES key out of the database & into a file. At least then you'd need to compromise the server and the database to decrypt passwords.

Keen to hear opinions on both of these, please?

johnnyq:
I know the AES key being stored in the database is definitely not the best of ideas. I like option 2 where we encrypt the AES key for each user. All users should be using good passwords anyway. I believe this is how big password managers like Bitwarden do it.

wrongecho:
#265 resolves this issue.
As a summary:
- At setup, a "master" AES key is generated and is encrypted with the initial user's password. This is done using PBKDF2 and a salt to generate a 128-bit key. AES-128-CBC (+ IV) is then used to generate the encrypted password. This is the "user specific key".
- Upon login, the user's password is used to decrypt their Specific Key in order to retrieve the master key. 
[...]

PR #265, opened 2022-01-10
Encryption changes

wrongecho:
Further to our discussion in #264, this pull changes the way that login entries are encrypted in the database.

At setup, a "master" AES key is generated (like before), only now it is encrypted with the initial user's password. Specifically, this is done using PBKDF2 and a salt to generate a 128-bit key (as passwords may be more or less than 16 chars). AES-128-CBC with an IV is then used to generate the encrypted password.

For example, my password of "itflow" is stored as: "$2y$10$dPiyQeQIajgGn/ZZwO30pOwExu6pKNPN4.ON5XsXgovslC2hzDqDS" in the database (the first 32 characters are the salt & iv).

Upon login, the user's password is used to decrypt their Specific Key in order to retrieve the master key.
I didn't want to store the master key in plaintext in the database or in the session (disk) so we now encrypt the master key again (with the users "session key") and store the ciphertext in the users session and the encryption key as a client cookie. 
In theory, this means that only this particular user will be able to access the master key by providing the cookie in a HTTPS request.
As soon as the user logs out, the session is destroyed and the cookie is unset. All that is left at that point would be the encrypted Specific key in the database.

All logins are encrypted using the master key and an IV to ensure that each login is secure.

The master key never hits the server disk. The only time it is shown to the user is at http://127.0.0.1/itflow/settings-backup.php to allow it to be backed up. This still probably needs a little bit of work to restrict this based on user role?

I think these changes should cover everything needed for this to work initially. This will require a reinstall to get things set up properly. The main functions are included in the functions.php file.

Please let me know your thoughts? :)

### DIFF
--- a/client_assets.php
+++ b/client_assets.php
@@ -39,7 +39,7 @@ if(isset($_GET['o'])){
 //Rebuild URL
 $url_query_strings_sb = http_build_query(array_merge($_GET,array('sb' => $sb, 'o' => $o)));
 
-$sql = mysqli_query($mysqli,"SELECT SQL_CALC_FOUND_ROWS *, AES_DECRYPT(login_password, '$config_aes_key') AS login_password FROM assets LEFT JOIN contacts ON asset_contact_id = contact_id LEFT JOIN locations ON asset_location_id = location_id LEFT JOIN logins ON login_asset_id = asset_id
+$sql = mysqli_query($mysqli,"SELECT SQL_CALC_FOUND_ROWS * FROM assets LEFT JOIN contacts ON asset_contact_id = contact_id LEFT JOIN locations ON asset_location_id = location_id LEFT JOIN logins ON login_asset_id = asset_id
   WHERE asset_client_id = $client_id 
   AND (asset_name LIKE '%$q%' OR asset_type LIKE '%$q%' OR asset_ip LIKE '%$q%' OR asset_make LIKE '%$q%' OR asset_model LIKE '%$q%' OR asset_serial LIKE '%$q%' OR asset_os LIKE '%$q%' OR contact_name LIKE '%$q%' OR location_name LIKE '%$q%') 
   ORDER BY $sb $o LIMIT $record_from, $record_to");
@@ -178,7 +178,7 @@ $num_rows = mysqli_fetch_row(mysqli_query($mysqli,"SELECT FOUND_ROWS()"));
 
             $login_id = $row['login_id'];
             $login_username = $row['login_username'];
-            $login_password = $row['login_password'];
+            $login_password = decryptLoginEntry($row['login_password']);
       
           ?>
           <tr>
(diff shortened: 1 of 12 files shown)


johnnyq:
This looks fantastic! your cryptography skills are beyond mine but your explanation is spot on and makes sense. This is a huge leap over the existing way passwords were stored and should make IT Techs feel more comfortable to store client password in ITFlow. Is it anyway possible to use the existing AES key for existing client logins and passwords. I guess im asking is there an upgrade path for this for users that have existing client password in their Database

johnnyq:
I like the backup Master Key. I'm wondering if we can have an option to import Master Key (Old AES Key) and it will just convert all the passwords over.

wrongecho:
Thank you! Glad you could make sense of my code/comments! I'm not a cryptography expert but this should be okay hopefully. :)

In terms of upgrading existing keys to use this format, this has been a bit of a nightmare to be honest hahaha... I've done some thorough testing and think it's ready now!!
Upgrades will be performed by a single user only. This user has to upgrade every company, and then reset all account passwords. 
I've tried to keep this whole thing as simple and user-friendly as possible. I understand the password reset requirement is annoying but it's a necessary evil due to the way this is set up (it's either that or we store user passwords plaintext...).
It's worth noting that you should do the upgrade procedure immediately after updating the files from git. Adding new login entries or app user accounts will break things.
The UI for the upgrade info is just text, but I figured as it's a one-time upgrade it doesn't need to be fancy.
Could you do some testing on your end please and let me know your thoughts?  :) 

Cheers!

Upgrade procedure:-
- Logout
- Backup existing AES key for each company
- Backup logins database, ideally the full DB
- Pull the changes to files made in this PR. No need to touch the database.
- Use HTTPS (or set $config_https_only = FALSE in config.php) (this affects cookies and was driving me in circles for about an hour!).
- Login as an admin user that has access to ALL companies on the installation
- Observe that logins can no longer be decrypted - the passwords don't show - this is because the system is trying (and failing) to use the new encryption method.
- Open /settings-update.php, review section "Update AES Key"
- Confirm that you can see the usernames and passwords (logins) that exist in the database as cleartext
- Enter your account password
- Click "Update encryption scheme for this company"
- Backup your new master key somewhere safe
- Repeat steps for each company

Testing:-
- Check for the presence of user_encryption_session_key cookie. 
- Test you can decrypt logins again
- Add a new password entry
- Log out
- Log back in
- Test that you can still decrypt logins
- Change your password, check you can still access logins (system will force a logout)
- Using this current admin account, set the passwords for other admins (This allows them to also have access to the master key)
- Login using the other admin account(s), check you can access logins
- Change password of other admin account(s), check you can still access logins

Other random thoughts:-
- May be worth mentioning in the setup document to backup the master key?
- We need to force the app to redirect HTTP to HTTPS, if available. Going to all this work is silly if the user is not using an encrypted connection in the first place.

johnnyq:
Worked like a charm @wrongecho !! This is Super Awesome! Thank you for your time on adding this and the in-place upgrade.

johnnyq:
1. I agree with mentioning it in the setup docs maybe,
2. I think forcing HTTPS should really be done on the web server configuration

wrongecho:
Happy to help, thanks for merging!

1. OK cool
2. Yeah, that's a fair point. Definitely more a role for the webserver to redirect. 

Thanks again!

PR #270, opened 2022-01-12
Ticketing updates

wrongecho:
- Fixed the delete ticket button deleting the client instead of the ticket(!)
- Hide edit/delete options when a ticket is in a closed state
- Added ticket merge


johnnyq:
1. lol Good catch @wrongecho I could see that being disastrous for an unsuspecting victim.
2. Makes for common sense good add
3. Fantastic!! I cant wait to try this one Thanks for this!!

PR #272, opened 2022-01-12
Global search

wrongecho:
- Fixed login decryption - need to add the copy and reveal button
- Fix client listing (ish)

I will work on the search tomorrow and link it into the main site search, etc.. Unless you had other plans for searching?

### DIFF
--- a/global_search.php
+++ b/global_search.php
@@ -6,10 +6,10 @@ if(isset($_GET['query'])){
 
   $query = mysqli_real_escape_string($mysqli,$_GET['query']);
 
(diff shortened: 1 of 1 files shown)


johnnyq:
I totally forgot about this function, lol. The initial intention for this feature was to be a real time search of just about everything (or at least the most crucial things) and list items separated by category headers. The criteria was to at least enter 3 or more characters before it did it's mass query and list up to 5 max per category header. The listing then would be clickable which would take you to the correct location of that element. I'm open to ideas as well and I trust yours.

johnnyq:
Global search was at one point activated by the top Search bar of the main Nav Bar. It should be relinked to that element.

PR #274, opened 2022-01-13
get_credentials.php

wrongecho:
Hello!

So despite planning to look at the search today, I started looking into browser extensions and then somehow spent most of the day developing my first ever Chrome/Firefox extension for ITFlow, lol.

The idea is that you login to ITFlow and then browse to a page that requires authentication that you have saved in ITFlow (like a firewall/router login page. etc). 
Assuming that you have the URL set correctly, you can click on the ITFlow extension which will send a request to _https://your.it.flow/get_credential?host=_ and add the site hostname. ITFlow will then send user/pass data to the extension, which will fill the username/password field on the page. 

This commit facilitates the framework for this. Further commits to come!
Hope you like the idea? :)

### DIFF
--- a/get_credential.php
+++ b/get_credential.php
@@ -1,0 +1,78 @@
+<?php
+// Headers to allow extensions access (CORS)
+$chrome_id = "to-be-confirmed";
+$firefox_id = "to-be-confirmed";
+$http_origin = $_SERVER['HTTP_ORIGIN'];
+if ($http_origin == "$chrome_id" || $http_origin == "$firefox_id")
+{
+    header("Access-Control-Allow-Origin: $http_origin");
+    header("Access-Control-Allow-Credentials: true");
+}
+
+include("config.php");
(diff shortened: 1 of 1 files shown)


johnnyq:
Wow!! What a super Surprise!! I love the idea!!  I cannot wait to try this. Technically this can now be used as full featured password manager!! One question where do I download the extension lol =]

wrongecho:
You're welcome!!

Really basic functionality for now but:
https://github.com/wrongecho/itflow-firefox
https://github.com/wrongecho/itflow-chrome

You'll need to install the extension in developer mode and then adjust the variables at the top of get_credentials.php to your temporary extension ID to allow the extension access (something like moz-extension://857479e9-3992-4e99-9a5e-b514d2ad0a82 for Firefox). 
Once the extensions are approved and assigned official IDs then we'll add these to the code and this won't be an issue.

Issue #275, opened 2022-01-14
Currency & Ticket System


crnbezuidenhout:
Awesome application/site btw...
I have downloaded the files and started playing around... one powerful application. 

I have noticed that when sending a invoice/quote via mail, the currency symbol stays as $ and not via the currency selected.
Also when creating tickets, the response or email does not get sent to the client.

wrongecho:
Welcome @crnbezuidenhout!
We're still in beta with the application at the moment so a lot of features are still in development or may have bugs. 
In terms of the e-mail currency not showing for quotes/invoices, I've just submitted a pull request to fix that, nice catch.
We still need to implement e-mail functionality in tickets. Outgoing emails should be relatively simple but processing replies is likely going to be more of a challenge and might be beyond the 1.0 release. We're also aiming to have a public frontend for the ticket system too eventually, I think.
If you come across any more bugs, feel free to raise an issue. Can't guarantee they'll be fixed right away but all feedback is appreciated! :)

crnbezuidenhout:
Thanks Johnnyq/Itflow

I will play around and will keep you posted should I find something.

Again this is a super awesome idea and I love the fact that it's invoplving
the community.

I have noticed the following so far:
- within tickets, its keeps showing "Needs Approval" on the right hand side
- The Invoice button does not do anything ( not a big deal as its still
beta )
- When a mail is sent, it states that the mail was sent however does not
refresh the page back to the invoices page, one has to go back manually

Thanks for the quick update and fix

Enjoy...

wrongecho:
I totally agree - I love the fact that this application is open source. It has such potential! 
On that note, if you find anything that you can fix yourself (even if it's just a typo, etc), feel free to submit a pull request. I'm sure Johnny appreciates all the help he can get.

I know that the "Needs approval" & Invoice button in ticketing was only recently added. I think the option is for break/fix hourly-billed clients so that if someone is not a primary/approved contact, work can't be done/invoiced until it's approved? Not entirely sure as it seems to say needs approval for all client contacts at the moment, (and as you say the invoice button is just a placeholder currently) but I'm sure they'll take shape over time.

Most form submissions will take you to the page you were on when you performed the action. Assuming this is working, do you mean that it doesn't go back to the main invoices page (/invoices.php) page? If so, I can see the appeal there. You've sent the invoice, you're done with that one, back to the main page to send another - kinda thought process?

crnbezuidenhout:
_

[...]

PR #276, opened 2022-01-14
Email currency

wrongecho:
From #275 - When sending an invoice/quote via mail, the currency symbol stays as $ and not via the currency selected.


johnnyq:
Nice Work!! I think this will need to be implemented in other places as well (cron.php other areas in post.php)

PR #277, opened 2022-01-14
Global search

wrongecho:
General edits
- Hyperlinking
- Added ability to search for tickets
- Adjusted navbar search to use the global search

We can search for clients, vendors, products, logins, and tickets. Is there anything else you'd like to search for? 

_PS: The code formatting was driving me mad trying to add a bit for tickets and keep it in line with the two spaces as tabs so I just let it adjust the formatting. Hope that's okay :)_


johnnyq:
Awesome @wrongecho this will provide a nice quick way to get information without navigating all around Good workk!! =]

johnnyq:
I really like the grid work! it looks very nice!!

PR #283, opened 2022-01-15
Update client_certificate.php

aftechro:
- fetch expiry date 

Future to do: Show issuer - atm, shows country of issue but not the provider, maybe @johnnyq or @wrongecho can have another look here

### DIFF
--- a/client_certificates.php
+++ b/client_certificates.php
@@ -100,13 +100,28 @@ $num_rows = mysqli_fetch_row(mysqli_query($mysqli,"SELECT FOUND_ROWS()"));
             $certificate_expire = $row['certificate_expire'];
             $certificate_updated_at = $row['certificate_updated_at'];
             $certificate_public_key = $row['certificate_public_key'];
+            
+$url = "https://$certificate_domain";
+$orignal_parse = parse_url($url, PHP_URL_HOST);
+$get = stream_context_create(array("ssl" => array("capture_peer_cert" => TRUE)));
+$read = stream_socket_client("ssl://".$orignal_parse.":443", $errno, $errstr, 30, STREAM_CLIENT_CONNECT, $get);
+$cert = stream_context_get_params($read);
+$certinfo = openssl_x509_parse($cert['options']['ssl']['peer_certificate']);
+// echo "Name: ".$certinfo['name']."<br/>";
+// echo "Subject: ".$certinfo['subject']['C']." ".$certinfo['subject']['ST']." ".$certinfo['subject']['CN']."<br/>";
+// echo "Issuer: ".$certinfo['issuer']['C']." ".$certinfo['subject']['O']." ".$certinfo['subject']['CN']."<br/>";
+// echo "Version: ".$certinfo['version']."<br/>";
+// echo "Valid from: ".gmdate("Y-m-d",$certinfo['validFrom'])." to ".gmdate("Y-m-d",$certinfo['validTo'])."<br/>"; 
+         // $validFrom = date('Y-m-d H:i:s', $data['validFrom_time_t']);
+         // $validTo = date('Y-m-d H:i:s', $data['validTo_time_t']);
 
           ?>
+       
           <tr>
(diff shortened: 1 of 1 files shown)


wrongecho:
Heyy. Just some thoughts re certificates (#286, #285, #284):

I thought I added functionality to fetch the expiry date when the certificate was initially added in an earlier PR - https://github.com/johnnyq/itflow/pull/230? I think it was only added for certificate add (and not edit). Was this working for you?

I would just caution against removing the expiry/issuer from the database - I think we're better off grabbing that at certificate add/edit/update and storing the value in the database, rather than parsing the certificate on the fly every time. We also probably want to support users that want to enter the issuer/expiry manually and not add the public key? Also, I think cron relies on the expiry date being in the database at the moment.

I think for issuer we might be able to use use `$issuer = $publicKeyObj['issuer']['O'];`.

What do you think?

aftechro:
Hey wrongecho, 

Whatever works best to get the details, plese note I'm not a dev, I'm learning things these days and I'm excited that I managed to do stuff which of course I still need to understand them 100% and what's behind it
My logic behind is to - when domain is added, automatically fill in the Ssl cert from the domain- sure, ideal is to get the option for both automatic fetching as well as manual input
Was not aware you've done something on ssl as when I tested was not working so I though still needs work to do 




Sent from Yahoo Mail on Android 
 
  On Sat, 15 Jan 2022 at 12:15, ***@***.***> wrote:   


Heyy. Just some thoughts re certificates (#286, #285, #284):

I thought I added functionality to fetch the expiry date when the certificate was initially added in an earlier PR - #230? I think it was only added for certificate add (and not edit). Was this working for you?

I would just caution against removing the expiry/issuer from the database - I think we're better off grabbing that at certificate add/edit/update and storing the value in the database, rather than parsing the certificate on the fly every time. We also probably want to support users that want to enter the issuer/expiry manually and not add the public key? Also, I think cron relies on the expiry date being in the database at the moment.

I think for issuer we might be able to use use $issuer = $publicKeyObj['issuer']['O'];.

What do you think?

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

wrongecho:
Hey aftechro - I'm not a dev either and am still learning as I go! A good quarter (or more) of what I commit is heavily assisted through StackOverflow haha ;)

I absolutely love the idea of being able to automatically link certificates with domains, and also your ideas of being able to get the domain details (expiry, NS, etc) as well as the certificate(s) and pull them into ITFlow automatically. 

I think your logic is good! My only thoughts were just about storing the data (so we don't have to wait for the server to try and pull it each time the page loads - maybe a "refresh" button instead?) and giving people users options for manual entry (maybe the certificate is only accessible via a VPN or something). I might be wrong here as I've just skimmed the pull requests. I just wanted to give my thoughts :).. I'll let Johnny take a look and see.

johnnyq:
closing this pull request for now as we will take an alternative approach

aftechro:
Agree with both of you, we just need to find the best approach based on each experience. I played a bit with that code, and i was so happy that i succeeded to do such big thing in developing :d I`m sure with practice, in few months i can get better and better :D 

- my logic and of course, based on experience - certs should be pulled from domain section, and allow manual input for vpn or other domains associated with the client. 
- domain WHOIS would also be great to have with all the zone records that it`s useful for when moving ns servers or some major rollback, you would know what was there before. ITGlue has it as Revisions

image
image

and ssl

image
image

aftechro:
Agree with both of you, we just need to find the best approach based on each experience. I played a bit with that code, and i was so happy that i succeeded to do such big thing in developing :d I`m sure with practice, in few months i can get better and better :D 

- my logic and of course, based on experience - certs should be pulled from domain section, and allow manual input for vpn or other domains associated with the client. 
- domain WHOIS would also be great to have with all the zone records that it`s useful for when moving ns servers or some major rollback, you would know what was there before. ITGlue has it as Revisions

image
image


and ssl

image
image

wrongecho:
Opened #289 to track this.

PR #286, opened 2022-01-15
Post.php

aftechro:


johnnyq:
Hi @aftechro your doing well man! I know your still learning but you are learning very quick. I think we are going to take a different approach to this but with the functionality your are looking for. We are going to close this commit and the others for now. Keep up the good work!

PR #290, opened 2022-01-16
Extension updates

wrongecho:
Making some progress with this.

See if you can get this to work with the [Chrome extension](https://chrome.google.com/webstore/detail/afgpakhonllnmnomchjhidealcpmnegc/)? Waiting for Firefox to review my resubmit..
You'll just need to set the URL in the extension options & pin the extension. In ITFlow, login & enable the extension in the user profile. Then just navigate to a login page that you've set in Logins and click the extension icon.

Let me know how you get on? :)

Troubleshooting:
- Your ITFlow instance must use HTTPS with a valid SSL (or at least one trusted by the browser)
- URLs: check www vs non www
- Check cookies are being set
  - user_extension_key and user_encryption_session_key should both be set, and match the values in the database
  - They should be set with httponly, secure and samesite=none
- Check the browser console
- Check extension console (dev tools need to be enabled)

### DIFF
--- a/db.sql
+++ b/db.sql
@@ -1166,6 +1166,7 @@ CREATE TABLE `users` (
   `user_token` varchar(200) DEFAULT NULL,
   `user_avatar` varchar(200) DEFAULT NULL,
   `user_specific_encryption_ciphertext` varchar(200) DEFAULT NULL,
+  `user_extension_key` varchar(18) DEFAULT NULL,
   `user_created_at` datetime NOT NULL,
   `user_updated_at` datetime DEFAULT NULL,
   `user_archived_at` datetime DEFAULT NULL,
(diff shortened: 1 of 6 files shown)


johnnyq:
Oh this is fantastic!! Another surprise Ill test after merge

johnnyq:
It works like a charm!! Fantastic!! although there are a few urls it didn't work with but we can, just to name a few login.microsoftonline.com and dropbox. Maybe we can make a running open issue that lists some common sites that have issues but this is the start to really something awesome! Keep up the great work!

wrongecho:
Glad you got it working!!

I agree, tracking sites that it doesn't work with via an issue would be best.

Currently, the extension recognizes login forms by looking for a password field and then finding the text field just before it (as the username/email field).
Some sites (like Microsoft) do that annoying "split" login where you enter your username/email, click next, and get taken to another page for your password. I need to add support for this somehow. 
Not too sure why you couldn't get Dropbox to work - add it to the issue and can take a look at some point :)

We've proven the core functionality/capability is there now, shouldn't be too much work to fine-tune it in the near future.

PR #296, opened 2022-01-17
Quote/Invoice redirect

wrongecho:
As per #275, I agree it makes sense to direct users back to the quotes/invoices main page after sending out the email. _You've sent the invoice, you're done with that one, back to the main page to send another._


johnnyq:
Alright sounds like a logical thing adding it in Thank you!!

PR #297, opened 2022-01-17
Add currency to price

wrongecho:
Just some final touches surrounding displaying currency - loosely related to #275


johnnyq:
ooops forgot about replicating that, thank you!

Issue #302, opened 2022-01-19
On New Ticket: Add the asset field


Wikiphil:
Can we add the Asset field in the ticket open windows?

![new-ticket](https://user-images.githubusercontent.com/84292480/150188411-d5d2007f-b9d0-440d-bc25-294a73ac9e53.png)

We have to open the ticket to choose the asset after creation

![edit-ticket](https://user-images.githubusercontent.com/84292480/150188612-7cfabbec-dc19-4c5e-866f-e2b0e6d3da5d.png)

Wikiphil:
When you already are in a Client! :-) From the main screen, it's a bit harder to list all assets.

wrongecho:
The reason this wasn't added on the ticket creation popup / modal is because the popup then becomes too big with the extra line. The ticket edit has the ticket assigned to box removed on the edit view, so that makes room for it.

We could definitely add it, but it means you then have to scroll a bit for most screens. Open to this, see what Johnny thinks? 

Also, I want to add support for multiple assets and maybe suggest assets that are assigned to the ticket contact.. Eventually..

Wikiphil:
I Understand, Client is already switched by Contact depand on "New Ticket" location.

wrongecho:
Added this via #307.

PR #307, opened 2022-01-20
Ticket enhancement

wrongecho:
- Removed the "email updates", as this is being handled via private/public updates
- Add functionality to add asset at ticket creation, if on a client page (if client ID is in GET) (#302)


johnnyq:
Ill get this merged looks good thank you

PR #308, opened 2022-01-20
Document Tags

wrongecho:
- Make name required.
- Make delete box select2 so it's filterable


johnnyq:
Merging in Here it comes thank you for patches

Issue #312, opened 2022-01-21
Feature request: Login brute force protection


wrongecho:
Login brute force protection
A few ways we could do this, but by IP address would probably be perfectly fine. 
Could possibly use the existing logging - or just track failed login attempts in a new table and check that a particular IP hasn't failed more than X times in the last X mins before progressing the login request.

johnnyq:
Very good idea! Maybe also have a 30 Min unblock window jsut in case someone really forgot their password

johnnyq:
Works great so far! @wrongecho

PR #320, opened 2022-01-22
Add basic IP login brute force protection

wrongecho:
#312


johnnyq:
Oh yeah this is gonna be a real good one thank you!!

PR #329, opened 2022-01-29
Send email to client on public ticket update

wrongecho:


johnnyq:
Very much needed and well done on the body of the message thanks @wrongecho

PR #334, opened 2022-01-29
Scheduled tickets - remove email ticket updates option

wrongecho:
Remove email ticket updates option - this is controlled by ticket note visibility


johnnyq:
makes sense Thanks @wrongecho

PR #335, opened 2022-01-30
Asset make should not be required (e.g. VMs)

wrongecho:


johnnyq:
Good call =] Merging time

PR #337, opened 2022-01-30
Linked certificate domain should not be required

wrongecho:


johnnyq:
Absolutely correct, linked domain should not be required merging in

PR #343, opened 2022-02-03
API

wrongecho:
Wanting to create a API with create, read, update and delete endpoints for each module. This should allow us to have good coverage for each module and keep everything as simple as possible.

E.g. api/v1/assets/read.php  or  api/v1/contacts/create.php

### DIFF
--- a/api/v1/assets/read.php
+++ b/api/v1/assets/read.php
@@ -1,0 +1,66 @@
+<?php
+require('../validate_api_key.php');
+
+if($_SERVER['REQUEST_METHOD'] !== "GET"){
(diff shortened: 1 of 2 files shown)


johnnyq:
Wow nice work @wrongecho 
I have worked on an API Keys tables that may work nice with this. Each key will have a name, an optional expiration date and ACLs for what it can access. This will also provide nice logging.

wrongecho:
> Wow nice work @wrongecho 
> I have worked on an API Keys tables that may work nice with this. Each key will have a name, an optional expiration date and ACLs for what it can access. This will also provide nice logging. 

Sounds great! More structured and secure way of doing the API I think.

PR #358, opened 2022-02-13
Move network edit modal data processing to AJAX

wrongecho:
As per our recent chat, this serves as a proof of concept of only pulling part of the data from the database on the main page and, instead, showing the rest once the edit modal is displayed dynamically with AJAX.

My dev site page has three networks with a load of roughly 0.0040s. With these changes, that decreases to 0.002s. Whilst you probably wouldn't notice the difference on the networks page specifically, you probably would notice this for tickets or assets where much more information is being retrieved.


johnnyq:
nice work Ill pull this in, thats quite a bit of a difference in load times and tottally agree with you on the tickets

PR #362, opened 2022-02-17
Scheduled tickets - View for no tickets

wrongecho:
Scheduled tickets - Add nothing to see here (manual - non-paginated version)


wrongecho:
Needs moving to the proper pagination but will do for now.

PR #364, opened 2022-02-19
Add document sharing via link

wrongecho:
It is a little messy at the moment will add functionality to share files & logins also.

Let me know your thoughts?

### DIFF
--- a/client_contact_edit_modal.php
+++ b/client_contact_edit_modal.php
@@ -103,7 +103,7 @@
                     <option value="">- Location -</option>
                     <?php 
                     
-                    $sql_locations = mysqli_query($mysqli,"SELECT * FROM locations WHERE (location_archived_at > NOW() OR location_archived_at IS NULL) AND location_client_id = $client_id ORDER BY location_name ASC");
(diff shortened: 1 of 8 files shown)


client_contact_edit_modal.php
```diff
@@ -103,7 +103,7 @@
                     <option value="">- Location -</option>
                     <?php 
                     
-                    $sql_locations = mysqli_query($mysqli,"SELECT * FROM locations WHERE (location_archived_at > '$network_created_at' OR location_archived_at IS NULL) AND location_client_id = $client_id ORDER BY location_name ASC"); 
+                    $sql_locations = mysqli_query($mysqli,"SELECT * FROM locations WHERE (location_archived_at > NOW() OR location_archived_at IS NULL) AND location_client_id = $client_id ORDER BY location_name ASC");
```
johnnyq, line 106: the reason why I had location_archived_at > $network_created_at in edit is if someone archived that location but was still linked to contact it would still show that link this location_archived_at > NOW() is okay to use on newly created items but dangerous to use on edit items

johnnyq:
This is a feature that will sure be a winner, thanks @wrongecho

client_contact_edit_modal.php
wrongecho: This was just something I came across whilst working on this. I was getting an error and assumed it was just a typo. No worries.

PR #368, opened 2022-02-20
Show OS if asset is a switch/router/fw

wrongecho:
I think that Switches, Routers, and Firewalls should still show the asset OS field. E.g. Cisco IOS


johnnyq:
okay I agree we'll add it on in

PR #369, opened 2022-02-20
Allow service notes to use multiple lines

wrongecho:


johnnyq:
I didn't even know nl2br() function existed haha. I implemented this a little differently when it came to show line breaks in invoice notes / quotes etc by using a style function below. This still allowed the use of strip_tags() function for extra security. I thought htmlentities was used for sanitizing output vars. Im not sure what is the right way though.

<div style="white-space:pre-line"><?php echo $invoice_note; ?></div>

wrongecho:
Yeah, I was googling "How to show lines PHP SQL" and it came up haha.

I'll change that back to strip_tags. I just wanted a way to be able to use < > / etc, not a huge deal, I'll look into it more another time - it doesn't seem to affect the line breaks.

PR #371, opened 2022-02-22
Make contact mobile default NULL

wrongecho:
UNless there is there a reason this was '''NULL'''?


johnnyq:
yeah it should have always been Im not sure what happened thanks =]

Issue #304, opened 2022-01-19
Certificate fetch functionality


wrongecho:
The newly added certificate fetch functionality (#299 / #289) seems to have an issue, and I'm not too sure how to approach it (thanks for the icon btw!). 

It seems like we load all of the current existing certificate modals and when adding a new one the javascript is first seeing the existing modals (as they are technically loaded but just hidden?) and grabbing the domain from there. I was only using the one certificate to test this so didn't run into it then.

I remember somewhere you've added numbering to the modal links, is that to prevent this type of issue? 
Maybe we could use PHP to insert the domain record ID into the HTML ID and call that with the javascript? 
Alternatively, it can be less fancy and when you click the fetch it just updates that particular record ID (via post/get). (But I do like the nice javascript!!)
Open to thoughts.

aftechro:
I seen on my end after fetching the cert and save the add_cert_modal, nothing happens. even though the alert message saying certificate was added. The certs added, not listed. not sure if this is mysql vs mariadb thing again

johnnyq:
Ill take a peek here

wrongecho:
Fixed as part of moving to dynamically loading modals - #373

PR #373, opened 2022-02-24
Certificate update

wrongecho:
- Dynamically load certificate edit modal with AJAX for faster page load times
- Also fixes certificate details fetch error (#304)


johnnyq:
nice Work! at @wrongecho

PR #383, opened 2022-02-26
Add functionality to pull/sync assets & info from MeshCentral

wrongecho:
I think you use MeshCentral... Could you give this a try and let me know your thoughts, please?

The only slightly annoying requirement here is that we need to install NodeJS on the server for this to work (along with 2 packages: minimist & ws). I didn't fancy rewriting what has already been written by the MeshCentral team into PHP whilst also having to learn WebSockets!

Once that's set up, populate your MeshCentral server details in /settings-general.php and configure your client group names to match. Hit the Sync button and it should link/create/update assets from MeshCentral.

meshcentral-settings

meshcentral-group

### DIFF
--- a/client.php
+++ b/client.php
@@ -34,6 +34,7 @@ if(isset($_GET['client_id'])){
     $client_net_terms = $config_default_net_terms;
   }
   $client_notes = $row['client_notes'];
+  $client_meshcentral_group = $row['client_meshcentral_group'];
   $client_created_at = $row['client_created_at'];
   $primary_contact = $row['primary_contact'];
   $primary_location = $row['primary_location'];
--- a/client_asset_edit_modal.php
+++ b/client_asset_edit_modal.php
@@ -44,7 +44,7 @@
                   <div class="input-group-prepend">
                     <span class="input-group-text"><i class="fa fa-fw fa-tag"></i></span>
                   </div>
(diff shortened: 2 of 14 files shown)


johnnyq:
Woah this is nice @wrongcho  This will be the first big integrated app nice work

wrongecho:
@johnnyq, glad you like it! It's fairly basic and doesn't sync a whole lot - MeshCentral records less than I thought it would... I'll probably re-write it in PHP eventually but I think it's a step in the right direction as far as integrations go! I really like the idea that information in ITFlow can be automatically kept up to date and synced across from other tools.

Do you think we should try and add support for paid RMMs/products too (if so, where to start?!)? I wanted MeshCentral to be the first as it's free and was on the roadmap in the readme, but I suspect not everyone will be as 100% open source as you are ;)

johnnyq:
We do use the Free and Opensource RMM tool called tacticalRMM which houses all the info we need like Hostname Make Model Serial MAC Address IP OS etc. This works with Meshcentral for remote desktop functionality. This would be awesome for Integration as well. 

For the future on integrations we can make them modular and I think the first step to that would be to create an integrations SQL table that links to other tables that take advantage of the integration for example assets, clients etc

But yeah I 100% agree this would be awesome but for now I think we should stick with Meshcentral and TacticalRMM

Unifi Controller would also make a nice integration as well.

So many possibilities

PR #392, opened 2022-03-08
Update tickets.php

aftechro:
Ticket subject to shown as popover when hover over the subject, for faster assignment :d


aftechro:
you can also have the ticket details at the buttom:
`    data-content="<?php echo $ticket_details; ?><hr><small>By: <?php echo $contact_name; ?> IN: <?php echo $client_name; ?>| 
    Created at: <?php echo $ticket_created_at; ?></small>">`

Issue #385, opened 2022-02-27
IP Spoofing


wrongecho:
In functions.php, the get_ip function first looks for several HTTP headers, and only if they are not present will it return the REMOTE_ADDR. 
```php
function get_ip() {
    $mainIp = '';
    if (getenv('HTTP_CLIENT_IP'))
        $mainIp = getenv('HTTP_CLIENT_IP');
    else if(getenv('HTTP_X_FORWARDED_FOR'))
        $mainIp = getenv('HTTP_X_FORWARDED_FOR');
    else if(getenv('HTTP_X_FORWARDED'))
        $mainIp = getenv('HTTP_X_FORWARDED');
    else if(getenv('HTTP_FORWARDED_FOR'))
        $mainIp = getenv('HTTP_FORWARDED_FOR');
    else if(getenv('HTTP_FORWARDED'))
        $mainIp = getenv('HTTP_FORWARDED');
    else if(getenv('REMOTE_ADDR'))
        $mainIp = getenv('REMOTE_ADDR');
    else
        $mainIp = 'UNKNOWN';
    return $mainIp;
}
```

In a typical setup with no WAF/reverse proxy, we should **only** be looking at the REMOTE_ADDR to find the client's IP address, all other variables should **not** be trusted.

In a setup with a reverse proxy, we should allow the app owner/admin to define what value to trust (e.g. with Cloudflare this would be _CF-Connecting-IP_).

In the current setup, I am able to spoof these headers and make the IP logged whatever I like. This has implications for auditing/logging as well as login brute force protection.

I think we should by default use REMOTE_ADDR, but allow the user to customize in the event the server is behind a reverse proxy.

![burp](https://user-images.githubusercontent.com/32306651/155884395-845a8141-9312-4017-a046-0d0307be77f6.png)

![ip-spoof](https://user-images.githubusercontent.com/32306651/155884360-80e261c3-6066-4322-be47-bfbd19154ff1.png)

johnnyq:
Nice find! @wrongecho HestiaCP does proxying by default. Nginx acts as a proxy / Static Asset Cache and passes data to Apache

wrongecho:
@johnnyq, been thinking about this one. I like the idea of defaulting to using REMOTE_ADDR but allowing the user to specify for their use case in case they are using something like HestiaCP, Cloudflare, etc. Do you think this should be a setting in the database (and UI) or in the config.php file?

johnnyq:
[...]

PR #399, opened 2022-03-20
Default get_ip to REMOTE_ADDR

wrongecho:
Related to #385 

This change defaults IP logging to using REMOTE_ADDR, with the option of setting it to use the forwarded for header by defining it in the config as `DEFINE("CONST_GET_IP_METHOD", "HTTP_X_FORWARDED_FOR");` 
I was unable to get this working without using a constant for some reason.. Looking at config.php, I think that the config app name, base url, https only and enable setup should probably all be constants as well as they should never change during execution.

Let me know your thoughts on this approach, please?

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -2,40 +2,40 @@
 
 function keygen()
 {
-    $chars = "abcdefghijklmnopqrstuvwxyz";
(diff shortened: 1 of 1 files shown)


johnnyq:
I agree that makes the most sense

PR #402, opened 2022-03-21
Update calendar_events.php

aftechro:
Added clickable URL to events


johnnyq:
oh kool idea I see they are using the event id but I can work with this. Thanks

PR #415, opened 2022-03-27
Updated login.php

aftechro:
Changed login page - same UI as admin logo


wrongecho:
@aftechro thanks for improving this, looks loads better!

PR #449, opened 2022-04-19
Replace deprecated <center> tag with CSS

wrongecho:
SonarCloud bugs.


johnnyq:
I had no idea <center> was deprecated but I think Bootstrap has a class for centering built in called text-center not sure how well it works though

wrongecho:
Ah you're right!! Looks a tad cleaner so will give it a go.

PR #454, opened 2022-04-30
Fixed SQL Query to remove duplicate rows

rjwil1086:
Added group by in SQL query to remove duplicate rows caused by joins.


wrongecho:
Heyy @rjwil1086, thanks for trying ITFlow!

I'm having a little trouble getting the assets page to duplicate assets/rows. Would you mind attaching a screenshot of what you're seeing to help me understand the bug this PR fixes, please?

johnnyq:
I see whats going on here. Assets and logins are supposed to be a SQL 1 to 1 relationship however under logins you may assign a login to an asset that already has a username and password tied to it which will cause the asset to be duplicated under assets.

group by clause won't be a proper fix for this.

Technically an asset can have multiple logins for example a file server  with 10 users on it.
The only way to fix this would be to create a many to many relationship which would involve creating another table
example
table: asset_logins
fields: asset_id, login_id

Im not sure how I want to approach this just yet ill sit on this one but leave this open for tracking

rjwil1086:
Sorry for the delay! Johnny figured it out though.

My SQL skills are pretty rusty. Group by was a quick/lazy fix. I'm happy to
run with making the database changes and submitting a new pull request.

I've previously used a pay-for solution (ITGlue). We would assign multiple
logins to the same asset often (root login, sudo user login, sometimes
software logins installed on the asset, etc).

All you have to do to create the bug is assign more than one login to an
asset.

[image: image.png]

johnnyq:
Closing this for now as this is corrected

PR #465, opened 2022-05-20
Misc changes

wrongecho:
- Fix overview ticket table
- Add internal note when creating an invoice from a ticket
- Remove an unused function
- Send a new e-mail to a tech when their account is created (if checkbox selected)


aftechro:
was trying to do the same: Add internal note when creating an invoice from a ticket, but good you sorted :D

PR #480, opened 2022-08-04
Update functions.php

jcpit:
Return IP if running from behind Cloudflare.


johnnyq:
Super!! Thanks @jcpit

Issue #497, opened 2022-11-14
Dashboard Financial Information 


HollowCyberSec:
I can't find a way to stop techs from seeing the financial information in the main dashboard. Any way to stop this?

wrongecho:
Hey @HollowCyberSec 
We plan to eventually have different dashboards. As a quick fix, for now, I'm proposing that we just redirect techs to the main client list page. Very little of the info on that financial dashboard would be relevant to their role.

PR #505, opened 2022-12-11
Role enforcement

wrongecho:
- Quick redirect fix for stopping techs seeing the financial dashboard. 
  - I know we eventually have plans to create more dashboards, so maybe we can show them automatically based on role?
  - This solves #497
- Remove the actions button on the clients.php page for techs/accountants (they weren't able to make edits anyway)


johnnyq:
Good Idea Excellent I like the redirection based off roles

PR #546, opened 2023-01-02
Rectified Error on Timer Ticket

chandachewe10:
When the administrator has created a ticket the timer always starts from 0 instead of starting counting from the 
time the ticket was created.
Also when the page is refreshed the timer gets back to zero and start counting again
instead of starting counting from the time the ticket was created. This pull request will show the actual duration which
has elapsed ever since the ticket was created and the timer will not be restarted when the page is refreshed.

### DIFF
--- a/ticket.php
+++ b/ticket.php
@@ -625,10 +625,36 @@ if($ticket_status !== "Closed"){ ?>
   <!-- Ticket Time Tracking JS -->
   <script type="text/javascript">
       // Default values
-      var hours = 0;
-      var minutes = 0;
-      var seconds = 0;
-      setInterval(countTime, 1000);
+     // var hours = 0;
+      //var minutes = 0;
+      //var seconds = 0;
+var startDateTime = new Date("<?= $ticket_created_at;?>"); // (start time and date from DB)
+var startStamp = startDateTime.getTime();
+
+var newDate = new Date();
+var newStamp = newDate.getTime();
+
+var timer; // for storing the interval (to stop or pause later if needed)
+
+function updateClock() {
+    newDate = new Date();
+    newStamp = newDate.getTime();
+    var diff = Math.round((newStamp-startStamp)/1000);
+    
+    var d = Math.floor(diff/(24*60*60)); 
+    diff = diff-(d*24*60*60);
+    var h = Math.floor(diff/(60*60));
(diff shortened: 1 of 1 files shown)


wrongecho:
Hey @chandachewe10. 

Thanks for the PR - we really appreciate your time and effort in looking into this! 

**Unfortunately, this isn't the functionality we're aiming for with time tracking.**

In the MSP space, it can be helpful to know how long you've worked on a ticket, for billing and efficiency purposes/metrics. The timer aims to help capture this by tracking the time you have the ticket page physically open (you can of course edit the value though). 
For instance, you may work on a ticket for 30 minutes before posting the first reply/note. The next day you work another 10 minutes on it and post another reply/note. You have therefore tracked around 40 minutes of your work day/time to this ticket.

It seems like your changes would make the timer work in such a way that it constantly/cumulatively counts up from the time the ticket was first created. In our example above, it would show as over 24 hrs of time tracked to the ticket. 
This functionality would be great for something like an SLA timer (which needs to run nearly all the time), but I think you may have misunderstood the purpose of the time tracking a little here :). 

I hope this explanation helps. Again, we really do appreciate your time and effort - please feel free to continue looking at the forum/issues/code for other potential areas of improvement 👍

chandachewe10:
@wrongecho so the purpose of the timer is to capture **_how long someone has worked on the ticket_** and **_not how long the ticket has been opened._** 

So the system does not capture how long the ticket has been in an open state. Do you consider adding a functionality
for the purpose of showing how long it took for the customer's issue to be attended to? As it only favors the person who is working on the ticket by capturing the time he/she has worked on the ticket for billing purposes.

I believe an open ticket basically suggests that an issue registered by a customer has not been solved yet. Do you consider 
adding the functionality of tracking how long it took for that issue to be resolved (open ticket) and probably
the ticket closing automatically after a certain duration of time if not attended to  (Like in most companies) so that the
end user can then re-establish contact on the same ticket if not attended to, say 3 to 5 days for example.

wrongecho:
Hey @chandachewe10

Yes, that's the purpose of the timer.

In the future we would be looking to add reporting on the time to resolve a ticket as part of the Reports module, I assume we'd use the ticket creation and closure dates for calculating this.

We wouldn't want an auto-close system as we don't support re-opening tickets.

PR #547, opened 2023-01-02
Misc changes from code security audit

wrongecho:
Code Security Audit (part 1)

As discussed, I've been reviewing all files in the codebase for (mainly) security-related issues.
Each commit has individual messages as to what changed - please let me know your thoughts :)


johnnyq (commented):
strip_tags are already done on the top and also in pagination_head.php

stripslashes was used during output because if you search for \\test then it would output \\\\test

Im going to revert this change only

PR #549, opened 2023-01-03
SHOWING REVENUES IN EACH PARTICULAR MONTH COLUMN 

chandachewe10:
So when you go under `Reports -> Income` to view the income summary, the revenues in each particular month are not shown except under the Total bottom row where they are added together with other incomes. This is because the code fetching the revenues in each particular month always picks an auto increment id . Check this code which picks the revenues in each particular month in `report_income_summary.php on line 65`:

 `$sql_revenues = mysqli_query($mysqli,"SELECT SUM(revenue_amount) AS revenue_amount_for_month FROM revenues WHERE revenue_id = $category_id AND YEAR(revenue_date) = $year AND MONTH(revenue_date) = $month");`
The code looks for the id of revenue_id instead of the revenue_category_id

### DIFF
--- a/report_income_summary.php
+++ b/report_income_summary.php
@@ -92,7 +92,7 @@ $sql_categories = mysqli_query($mysqli,"SELECT * FROM categories WHERE category_
                 $payment_amount_for_month = $row['payment_amount_for_month'];
 
                 //Revenues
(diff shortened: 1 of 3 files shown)


johnnyq:
Hi @chandachewe10 Thank you for the fixes. Especially the revenue one!
One of commits you have in this PR changes the way the countdown timer works for tickets. This functionality does not work with what we are trying to do as @wrongecho has stated. Can you please resubmit this PR without the countdown timer changes, so we may merge your changes. Thank you and happy New Year!

chandachewe10:
To be resubmited without the countdown timer

PR #553, opened 2023-01-07
removed warning

chandachewe10:
If you mistakenly enter the **wrong credentials (email address)** while login in, it was throwing the warning of "Trying to access array offset on value of type null". This is because on line `57 in login.php` there is this code:

`$row = mysqli_fetch_assoc(mysqli_query($mysqli, "SELECT * FROM users LEFT JOIN user_settings on users.user_id = user_settings.user_id WHERE user_email = '$email' AND user_archived_at IS NULL"));
if (password_verify($password, $row['user_password'])) {......`

But if the email does not exist (wrong credentials) then `$row will become NULL` and the password verify will compare the input password with NULL which will throw a warning. Inserted a condition to prevent password verify if the email does not exist.


johnnyq:
Great Thanks @chandachewe10

PR #556, opened 2023-01-09
Moving docker $repo_branch var to config.php

lued:
This is to avoid changing settings_update.php post docker entrypoint.sh manipulation. Instead have all variable changes occur within config.php for consistency.


johnnyq:
@lued Excellent looks good! however we will also need to add the variable to existing installs configs. This should happen in the update function to append the repo branch to existing config.php versions.

johnnyq:
@lued we do have a database upgrade function we may be able incorporate the additional config var with this function, let me us do some brain storming before we pull the request in.

johnnyq:
@lued We figured out a solution we are using the database update function to update config.php for the repo_branch var seems to work going to pull this in

PR #559, opened 2023-01-18
Re-add login fix from chandachewe10

wrongecho:
Re-adding the login fix from #553 but with a slightly different implementation. This prevents a null array error when a user that doesn't exist tries to authenticate (@chandachewe10 - thank you!).
Also minor tidying.


wrongecho:
_Can probably ignore the SonarCloud vuln/issue for the most part. A user would only be able to XSS themselves if their password was HTML code... Encoding HTML characters wouldn't help as then the password would fail auth after they enter the 2FA code._

PR #570, opened 2023-01-23
Testing adding a background script to keep sessions alive

wrongecho:
Add a background bit of javascript that loads with most pages. This will call ajax.php (with no params) which should "renew" the login session and ensure it doesn't get garbage collected (though I'm still not entirely sure why that happens - unable to reproduce myself).

Related to https://forum.itflow.org/d/85-ticket-creation-box-allowed-when-auto-logged-out and https://forum.itflow.org/d/71-changing-auto-logoff-time

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -7,11 +7,11 @@
 	<!-- ./wrapper -->
 
 	<!-- REQUIRED SCRIPTS -->
-	
+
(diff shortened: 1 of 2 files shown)


wrongecho:
Closing as discussed @johnnyq - issue actually seems to be a privacy plugin.

PR #597, opened 2023-02-01
Re-add domain refreshing to cron

wrongecho:
Add domain info (Expiry, WHOIS, A, NS, MX, TXT) refreshing - one domain per day (sorted by oldest updated).

I know this was previously [removed](https://github.com/itflow-org/itflow/commit/dbea44591c9590b1068c455ec09aad5a72d95767) due to it causing issues with invoice/reminder emails.

I've added the company_id attribute this time (even though it's soon to be removed) as I think that was the issue before. 

I've done some testing on my dev instance and everything seems to be working as far as I can see. Could you please do some testing on your end and let me know?

Also worth noting I've converted cron to 4 spaces now, instead of 2.

Thanks!

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -412,7 +412,7 @@ while($row = mysqli_fetch_array($sql_companies)){
 
             } //End if Autosend is on
         } //End Recurring Invoices Loop
-    
+
         if($config_telemetry == 1){
 
             $current_version = exec("git rev-parse HEAD");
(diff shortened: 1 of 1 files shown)


wrongecho:
Test these changes at: https://cronreaddrefresh597.pr-review.itflow.org  
 _(automatic message)_

PR #606, opened 2023-02-05
API Updates

wrongecho:
- Add PHP logging when API queries fail because of SQL errors
- Add user agent to logging
- Enhance asset update endpoint


wrongecho:
Test these changes at: https://api606.pr-review.itflow.org  
 _(automatic message)_

PR #616, opened 2023-02-11
Add read api endpoint for payments

wrongecho:
This will allow retrieving all payments, payments via a payment ID, and (probably the most useful) payments against an invoice ID.

PS example: https://github.com/itflow-org/itflow-api-powershell/blob/main/payments/read.ps1


wrongecho:
Test these changes at: https://apipayments616.pr-review.itflow.org  
 _(automatic message)_

PR #617, opened 2023-02-11
POST 'models'

wrongecho:
Deduplicate POST value assignment for add/edit operations in post.php with a 'model' concept. This should also help prevent things breaking when one of the functions are updated but the other is missed.


wrongecho:
Test these changes at: https://postinputmodels617.pr-review.itflow.org  
 _(automatic message)_

PR #618, opened 2023-02-12
Swap CURDATE for NOW

wrongecho:
In post.php, we should use NOW() to track the date & time a notification was dismissed. CURDATE() only tracks the date.


wrongecho:
Test these changes at: https://dismissnotifdatetime618.pr-review.itflow.org  
 _(automatic message)_

PR #619, opened 2023-02-12
Adjust core files to 4 spaces

### LINKED ISSUE #538: Code Standards: Spacing
We should decide on whether we're using 2 or 4 spaces, and enforce this consistently across the codebase. 
Also need to update the relevant docs.

**Edit: After further discussion we are going to follow the [PSR](https://www.php-fig.org/psr/psr-12) recommendation of 4 spaces. **

To-do: Convert all code/files to 4 spaces instead of 2

wrongecho:
This should be the bulk of files moved to 4 spaces.
Might have missed a few files, but they should get caught when we next edit them. Closes #538

### DIFF
--- a/blank.php
+++ b/blank.php
@@ -2,10 +2,10 @@
 
 <!-- Breadcrumbs-->
 <ol class="breadcrumb">
-  <li class="breadcrumb-item">
-    <a href="index.html">Dashboard</a>
-  </li>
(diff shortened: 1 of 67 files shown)


wrongecho:
Test these changes at: https://spacing619.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Fantastic!! I like this we should deff implement this across the board

PR #621, opened 2023-02-12
Move upload checking to a function

wrongecho:
If we like how this works, we can convert all other uploads to use this.

You pass checkFileUpload() the file (from the superglobal $_FILES like `$_FILE['myfile']` along with an array of allowed extensions. It will check the file size is OK and the extension is allowed. If both are OK, it will return a sanitized file name.


wrongecho:
Test these changes at: https://files621.pr-review.itflow.org  
 _(automatic message)_

PR #625, opened 2023-02-14
Require CSRF when enabling/disabling users

wrongecho:
- Require CSRF when enabling/disabling users
- Small tidy of users.php
- Add deprecation notice to the companies adjust modal (we should probably aim to allow per-client access instead in the future!)


wrongecho:
Test these changes at: https://users625.pr-review.itflow.org  
 _(automatic message)_

PR #626, opened 2023-02-24
MySQL Compatibility

taynted:
MySQL doesn't like zeros in dates.

Converted
-  0000-00-00 to 1000-01-01
- 9999-00-00 to 9999-01-01


wrongecho:
_Related (for anyone following): https://forum.itflow.org/d/103-mysql-compatibility-fix_

johnnyq:
Closing as the dates search function has been merged into 1 file pagination_head.php instead of each individual file

PR #634, opened 2023-02-26
Add bulk delete option for scheduled tickets

wrongecho:
This PR adds the ability to select and bulk delete scheduled tickets. 

We'll need to tidy up the UI a little. Ideally we'd have this option under a "Bulk Actions" menu or something (with other options like duplicate, reassign, etc). 

We'll be able to replicate this for other areas of the app once we've got it looking/working how we want. An issue we'll run into is that we haven't converted all modals to use ajax - some modals are generated 10 times with the data instead of being dynamically populated. As you can't have a form inside a form, we'd need to fix this first.

image


wrongecho:
Test these changes at: https://bulkupdate634.pr-review.itflow.org  
 _(automatic message)_

johnnyq (commented):
I don't think these vars need to be defined and checked because using intval automatically sets them to 0 if not defined

so all we would need in one example is just this line of code` $asset_id = intval($_POST['asset'];`

johnnyq:
This is going to be amazing Fantastic Work @wrongecho but your right we are going to need to convert all the edit modals to dynamically update. Ideally would like the bulk actions button to appear on the top right before new scheduled ticket or just below new ticket in card body

wrongecho:
> I don't think these vars need to be defined and checked because using intval automatically sets them to 0 if not defined
> 
> so all we would need in one example is just this line of code` $asset_id = intval($_POST['asset'];`

Yes, intval sets it to zero but first checking if the vars are defined seems to prevent the PHP undefined error when trying to assign a variable that hasn't been passed:

Depending on if you're on the main scheduled ticket page or the client scheduled ticket page, a slightly different HTML form is shown. If you create a new scheduled ticket from the main page, there's no option for client contact or asset as the ticket could be for any client.

wrongecho:
> This is going to be amazing Fantastic Work @wrongecho but your right we are going to need to convert all the edit modals to dynamically update. Ideally would like the bulk actions button to appear on the top right before new scheduled ticket or just below new ticket in card body

Thank you - lots more work to go! I will leave the styling to you as you have a better eye than me - feel free to move it around, etc. I like the idea of it being somewhere in the card body, not the header.

johnnyq:
> option for client contact or asset as the ticket could be for any client.

I see that makes sense @wrongecho

PR #651, opened 2023-04-08
Bufix: Fix client file uploads

wrongecho:
The recent(ish) move to checking file uploads via a function means we broke the automatic (friendly) naming of files and the image previews in the on client_files.php.

This PR adds logic to grab the file extension and set the file (friendly/visible) name to that of the uploaded file, if one is not specified during the upload (as it's optional).

---

image

image

image


wrongecho:
Test these changes at: https://fixfileuploadautoname651.pr-review.itflow.org  
 _(automatic message)_

PR #652, opened 2023-04-08
Client file deletion

wrongecho:
- Add file delete confirmation modal
- Change file deletes to POST not GET requests in post.php as they are state changing
- Require CSRF validation for file deletes

(#181)


wrongecho:
Test these changes at: https://confirmfiledelete652.pr-review.itflow.org  
 _(automatic message)_

Issue #514, opened 2022-12-17
Feature Request: Ticket metric reporting


wrongecho:
On the back of #225 - "_Add more to reporting (Average Response and Resolution Time, Service Requests, Ticket satisfaction)_"

- Ticket counts (e.g. X tickets opened this month)
- Average response time (need to determine if a "response" is a ticket being assigned or replied to)
- Average resolution time
- Ticket satisfaction stats
  - We already have the ticket satisfaction smiley/sad faces - with notifications issued if a sad face is selected.

wrongecho:
Halfway there
![image](https://github.com/itflow-org/itflow/assets/32306651/6d8709b3-fc84-4c83-b482-e66638280b24)

wrongecho:
I think we need to add a 'ticket_first_response_at' field to ticketing to make this achievable, otherwise we're having to parse every single reply each time report loads. This also makes any sort of first response SLA MUCH more realistic.
This could be automatically be set when a reply is added to a ticket. If null, we could add a visual cue to a ticket to show the first response is outstanding (red asterisk)?

PR #656, opened 2023-04-10
Ticketing Updates - Change client

wrongecho:
This PR adds the ability to change a ticket's client

image

Additionally we're adding a new database field: `ticket_source` and will soon retire `ticket_created_by`. 
- Ticket source will show where a ticket was raised from: an agent, the portal, via email, scheduled ticket, etc. Once this is implemented across the code base, we could also report on it to show the breakdown of where your tickets are raised (#514).
- Ticket created by was originally designed when contacts were considered users.  It quickly became clear that it was better to separate contacts and users, meaning there is now no way to differentiate whether the value in ticket_created_by is a user or contact. Given that we track a tickets assigned agent and client contact, and log all tickets raised, it seems safe to remove this field - for the sake of simplicity.

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -345,6 +345,45 @@ if (isset($_GET['quote_get_json_details'])) {
 }
 
 /*
+ * Returns sorted list of active clients
+ */
+if (isset($_GET['get_active_clients'])) {
+
+    $client_sql = mysqli_query(
+        $mysqli,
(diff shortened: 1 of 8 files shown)


wrongecho:
Test these changes at: https://ticketing0423656.pr-review.itflow.org  
 _(automatic message)_

PR #666, opened 2023-04-30
Send invoices to all billing contacts + primary

wrongecho:
When manually raising an invoice, send copies to all billing contacts - not just the primary contact.

We should probably add this to cron as well.

https://forum.itflow.org/d/150-invoice-send-to-email-address


wrongecho:
Test these changes at: https://emailinvoice666.pr-review.itflow.org  
 _(automatic message)_

PR #674, opened 2023-05-11
Fix "My Tickets" button CSS

bhopkins0:
This pull request fixes the CSS for the "My Tickets" button.

Before:

[](https://postimg.cc/SnbZvCkX)

After:

[](https://postimg.cc/Whwrsk0q)


johnnyq:
Wow that looks much better it has always bothered me before Thanks for the fix @bhopkins0

Issue #346, opened 2022-02-05
Client Tags not searchable within client search


johnnyq:
I need to find a way to join client tags to the client query currently you cannot search by tags under client listing.

**Table Sample**

- **tags** [ tag_id, tag_name ]
- **client_tags** [ tag_id, client_id ]
- **clients** [ client_id, client_name ]

Possibly use a mix of **MySQL UNION** and **GROUP_CONCAT()** AS tags to group all the tags under one field called tags separated by commas which then can be used in a WHERE Clause example: `WHERE tags LIKE '%$query%'`
not sure where or how to use them

**Current Client Query**

```
$sql = mysqli_query($mysqli,"SELECT SQL_CALC_FOUND_ROWS * FROM clients 
  LEFT JOIN contacts ON clients.primary_contact = contacts.contact_id AND contact_archived_at IS NULL
  LEFT JOIN locations ON clients.primary_location = locations.location_id AND location_archived_at IS NULL
  WHERE (client_name LIKE '%$query%' OR client_type LIKE '%$query%' OR contact_email LIKE '%$query%' OR contact_name LIKE '%$query%' OR contact_phone LIKE '%$query%' 
  OR contact_mobile LIKE '%$query%' OR location_address LIKE '%$query%' OR location_city LIKE '%$query%' OR location_state LIKE '%$query%' OR location_zip LIKE '%$query%') 
  AND DATE(client_created_at) BETWEEN '$date_from' AND '$date_to'
  AND clients.company_id = $session_company_id $permission_sql
  ORDER BY $sortby $order LIMIT $record_from, $record_to
");
```

Issue #190, opened 2021-09-09
Feature: Client Tags


johnnyq:
This feature will allow you to add multiple tags to your clients which will be searchable. This will in turn will expand / replace the Support Field Maintenance and non-Maintenance which will be tagged feature. Common Tags could include Maintenance, non-Maintenance, Web Hosting, Network monitoring etc. These tags can be customizable

johnnyq:
DB Structure Added

johnnyq:
d23d455cd9ee9c86c3e16ceb0d1f318e95620806

johnnyq:
still need to be able to search client tags within the client listing not sure the best way to implement this. Also Display up to three tags and add a line break so tags don't run off

wrongecho:
Related to #346

johnnyq:
closing this one as client tags are complete

PR #675, opened 2023-05-11
Fixed clients.php blank page

bhopkins0:
# Issue

On a fresh install (from today) of ITFlow, I could not figure out how to add a client. It turned out that there was a MySQL error caused from a query in the `clients.php` files.

Here is a screenshot of the problem with PHP error reporting enabled:

[](https://postimg.cc/d7bCxdcz)

# Fix

I fixed this by by adding `client_tags.client_tag_tag_id` to the `GROUP BY` portion of the query. I believe you could also fix the error with this MySQL query: `SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));`. However, I felt it was better to modify the original query than set a global variable.

# Testing

The modification in created in this pull request fixed the issue. I was then able to create a client, create a tag, add a tag to a client, and modify the client information afterwards. Since I did not modify the query in any fundamental way, I do not believe it would cause break any other part of the code.

[](https://postimg.cc/dknytfyK)

### DIFF
--- a/clients.php
+++ b/clients.php
@@ -20,7 +20,7 @@ $sql = mysqli_query(
     OR contact_mobile LIKE '%$phone_query%' OR location_address LIKE '%$q%' OR location_city LIKE '%$q%' OR location_state LIKE '%$q%' OR location_zip LIKE '%$q%' OR tag_name LIKE '%$q%' OR client_tax_id_number LIKE '%$q%')
     AND client_archived_at IS NULL
     AND DATE(client_created_at) BETWEEN '$dtf' AND '$dtt'
-    GROUP BY clients.client_id
+    GROUP BY clients.client_id, client_tags.client_tag_tag_id
     ORDER BY $sb $o LIMIT $record_from, $record_to
 ");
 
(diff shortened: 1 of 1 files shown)


johnnyq:
Hi @bhopkins0 Nice work however the problem we ran across was with adding client_tags.client_tag_tag_id
It duplicates, triplicates the clients depending the number of tags attached to the clients. For example create another tag and assign it to the same client, that client will now duplicate in the client listing.

johnnyq:
see #190 and #346 for details
along with these forum posts
https://forum.itflow.org/d/115-clients-page-stopped-displaying-again/6

johnnyq:
We have been having a heck of a time cracking the code, but if you can solve this puzzle, i'm pretty sure you will be a hero to anyone that uses MySQL ;-)

bhopkins0:
Just created a new commit that should fix it, including the duplicates problem. 

I have tested and had no issues with: 
* Searching with tag name, company name, company industry, primary contact name
* Adding new clients
* Adding multiple tags to clients

Please let me know if there are any issues you encounter while testing it.

[](https://postimg.cc/vgBYKvtF)

johnnyq:
impressive @bhopkins0 we have been trying to tackle this one for a long time now. The commit looks great we will be reeling it on in.. Great Work!!

Issue #591, opened 2023-01-31
Bug: PHP 8.1 deprecates passing null values to htmlentities, preg_replace, numfmt_format_currency, etc.


wrongecho:
PHP 8.1 has deprecated the ability to pass null parameters to htmlentities, preg_replace and other functions.

![Image](https://user-images.githubusercontent.com/32306651/215805985-cd1e3e39-89cd-4e45-8374-375cecf3c562.png)

- https://php.watch/versions/8.1/internal-func-non-nullable-null-deprecation
- https://stackoverflow.com/questions/71707325/migration-to-php-8-1-how-to-fix-deprecated-passing-null-to-parameter-error-r

johnnyq:
Well that's a mess. Hmm I don't understand why php will not accept "NULL" values  when passing functions to them. I think PHP foundation should rethink this one. Theres no good reason for this.

johnnyq:
impressive @bhopkins0 we have been trying to tackle this one for a long time now. The commit looks great we will be reeling it on in.. Great Work!!

wrongecho:
Seems like @bhopkins0 fixed this with a custom function. Please re-open if this isn't fully fixed.

PR #676, opened 2023-05-11
Add function to use htmlentities without deprecated error

bhopkins0:
# Issue
PHP 8.1+ gives errors about `htmlentities()` if null is provided as the argument. This pull request will make it easy to solve issue https://github.com/itflow-org/itflow/issues/591.

Here is an example of clients.php using `htmlentities()` with error reporting on:

[](https://postimg.cc/2bhwPgdg)

# Fix

This pull request creates a function called `nullable_htmlentities()` in functions.php.  The function is the same as `htmlentities()`, however it will replace null values with any empty string.

I believe this function is the best method of fixing the deprecation error because it is easier to CTRL + R and replace all instances of `htmlentities()` than it is to manually add `?? ''` to each argument in all instances of `htmlentities()`.

Here is an example of clients.php using `nullable_htmlentities()` with error reporting on:

image

Please note that this pull request only includes the function, and does not replace any `htmlentities()` functions in the codebase.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -45,6 +45,10 @@ function key32gen()
     return $key;
 }
 
+function nullable_htmlentities($unsanitizedInput) {
+    return htmlentities($unsanitizedInput ?? '');
+}
+
 function initials($str) {
     if (!empty($str)) {
         $ret = '';


wrongecho:
Test these changes at: https://patch1676.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
@bhopkins0 Awesome!! Ill go ahead and start replacing the htmlentities() with the new function you created for us nullable_htmlentities()

PR #677, opened 2023-05-12
Fix domain registrar/webhost update not working

wrongecho:
https://forum.itflow.org/d/164-unable-to-update-any-domains

- Quote the expiry field when updating a domain
- Better handle no web host being set


wrongecho:
Test these changes at: https://domainexpirequotes677.pr-review.itflow.org  
 _(automatic message)_

PR #681, opened 2023-05-13
Cron - Send invoices to all billing contacts + primary

wrongecho:
Follow on from #666 - Invoices generated via cron should be sent to all billings contacts, rather than just the primary

https://forum.itflow.org/d/150-invoice-send-to-email-address


wrongecho:
Test these changes at: https://cronemailinvoice681.pr-review.itflow.org  
 _(automatic message)_

PR #683, opened 2023-05-14
Ticket.php

wrongecho:
- Change wording of ticket 'Respond' button to 'Add note' if reply is internal
- Fix TinyMCE not loading under certain circumstances
- Correct varname 'prefix' to 'ticket_prefix' in reply alert message


aftechro:
my next project was to add tabs for Post reply and Internal note, but i`m stuck with functions behind

image

Issue #684, opened 2023-05-15
client homepage - wrong href ("//")


Heart1010:
https://github.com/itflow-org/itflow/blob/eb4c5cbf344d2c2c128d5ae1329e1c860ed69810/inc_client_top_head.php#L19

The two slashes ("//") here in the href tag don't look correct. This produces a wrong/not working client URL (e.g. ```https://https//www.xy....```)

wrongecho:
Hey @Heart1010 

Thanks for getting in touch. I'm not sure exactly why it's coded this way but: "The 'two forward slashes' are a common shorthand for "request the referenced resource using whatever protocol is being used to load the current page" ([src](https://stackoverflow.com/a/9646435)).

Are you able to verify on the demo that this actually breaks the hyperlinks? This seems to function fine for me.
![image](https://github.com/itflow-org/itflow/assets/32306651/4a7967a2-ef10-4f80-8176-a1fe3dfbbb10)

Heart1010:
Test it please when you store the homepage url of the client with https://www....

![image](https://github.com/itflow-org/itflow/assets/1481680/b28fb59d-748d-4108-bd95-0ad04a6d196a)

wrongecho:
I see what you mean, but the prompt specifically leaves off the protocol:
![image](https://github.com/itflow-org/itflow/assets/32306651/95a72034-d604-403f-ad4f-70e7f4705be1)

Is there a specific reason you need to explicitly specify the http/https part?

Heart1010:
Simply when you copy/paste a domain/url from the browser or from an email signature and so on... it always will be added **with** the https:// part - and I think everyone is here copy and pasting the url and don't write it in per hand manually ;)

So I don't think it's a good idea here to *hope* that an admin puts in an url without the https:// part.
When there really shouldn't be https:// in at all - you have to make sure the https:// part is cut off or something else when storing it. *JM2C*

johnnyq:
This shoukd remove http:// and https:// in the rest of the code

9449a87a1f6747d017dedc085d94634755037b71

Thanks @wrongecho

PR #688, opened 2023-05-20
Add regex to remove http:// or https:// from client_website

wrongecho:
Issue #684

We only expect a domain in this field (e.g. example.com) but allow https://example.com. This breaks the link on the client banner and ticket parsing based on domain.


wrongecho:
Test these changes at: https://clientwebsiteremoveprotocolregex688.pr-review.itflow.org  
 _(automatic message)_

PR #690, opened 2023-05-28
Update ticket_invoice_add_modal.php

aftechro:
included the client_rate to be added automatically. Quantity to be changed as needed.


wrongecho:
Test these changes at: https://patch12690.pr-review.itflow.org  
 _(automatic message)_

wrongecho:
Looks good to me! Can we just fix the code smells please? 

Just a suggestion, but if we're pre-filling the hourly rate we may want to add something in the product field (like "Hourly Support"). 

It'd also be nice if we could work out the time tracking/rounding to automatically update the quantity, but I get that's a lot more work :)

aftechro:
**Looks good to me! Can we just fix the code smells please?**
_YES, now i can! thanks for the license_ 

**Just a suggestion, but if we're pre-filling the hourly rate we may want to add something in the product field (like "Hourly Support").** 
_That`s a good idea actually, the client would know what that will represents_

**It'd also be nice if we could work out the time tracking/rounding to automatically update the quantity, but I get that's a lot more work :)**
_Here i have an issue, i dont know how to pull this off... what we can do if i may, if tt < 30 minutes, charge as 30 minutes, if > 30 minutes charge one hour and so on....or whatever multiplies needed (10,15,30 - i suppose for this need additional invoice charge option).... we only use 30 minutes (as minimum charge),_

PR #691, opened 2023-05-28
Update invoices.php

aftechro:
guest view invoice fixed

### DIFF
--- a/portal/invoices.php
+++ b/portal/invoices.php
@@ -95,7 +95,7 @@ $invoices_sql = mysqli_query($mysqli, "SELECT * FROM invoices WHERE invoice_clie
                     ?>
 
                     <tr>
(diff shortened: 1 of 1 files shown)


wrongecho:
Test these changes at: https://patch13691.pr-review.itflow.org  
 _(automatic message)_

wrongecho:
Heyy! I just tried the demo and the link works fine - what's the issue here? Any changes made should be replicated over to portal quotes.php too.
_Does your URL set in config.php have a trailing slash by any chance?_

aftechro:
when clicked on invoice, it will give me file://url/url/guest_view.... so i`ve removed the //base_url/portal and just changed to ../guest_view... and works fine 

not sure if this is just an NGINX users issue, or this could be the config_base_url not being set with https:// in front. Strange is that when hover over the INV-number i get that format file://url/url/portal/guest_view_invoice.....

wrongecho:
@aftechro - Does this work correctly on the demo for you?

If it helps: My config_base_url is `127.0.0.1/itflow` on my Windows dev and just `demo.itflow.org` on the demo. What's yours?

johnnyq:
Hi @aftechro the proper way would be to use // before it was \\ which causes windows edge to think it was a file share.

PR #694, opened 2023-06-03
Enable login key code

wrongecho:
_Related to #680_

This is an optional feature.

When enabled (Settings > Security), you are required to provide your chosen key in the URL like `demo.example.com/login.php?key=mykey` in order to access the login page. 
This adds a little additional security to the login page, and allows a smooth transition for users to get to the client portal.

Behaviour:
- Login key disabled: Nothing changes (default)
- Login key enabled & correctly provided in URL: Login page is shown as normal
- Login key enabled & not provided/incorrect: Browser redirects to the client portal login page

### DIFF
--- a/login.php
+++ b/login.php
@@ -46,18 +46,18 @@ $config_smtp_password = $row['config_smtp_password'];
 $config_mail_from_email = $row['config_mail_from_email'];
 $config_mail_from_name = $row['config_mail_from_name'];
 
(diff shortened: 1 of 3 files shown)


wrongecho:
Test these changes at: https://enableloginkey694.pr-review.itflow.org  
 _(automatic message)_

PR #695, opened 2023-06-03
Client Overview - Enhance Recently Updated section

wrongecho:
- Move the asset warranty expiring and stale ticket queries out of inc_all_client.php and into client_overview.php, there is no need to load this data on every page load - just for the overview
- Show recently updated documents (in addition to contacts/vendors)
- General tidy

### DIFF
--- a/client_overview.php
+++ b/client_overview.php
@@ -2,19 +2,69 @@
 
 require_once("inc_all_client.php");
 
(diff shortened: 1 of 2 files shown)


wrongecho:
Test these changes at: https://recentlyupdated695.pr-review.itflow.org  
 _(automatic message)_

PR #699, opened 2023-06-17
Login page enhancements

wrongecho:
- Default to secure cookies (in case var is not defined in config.php)
- Enable content security policy
- Return HTTP 401 response code for invalid username/password combinations


wrongecho:
Test these changes at: https://loginenhancements699.pr-review.itflow.org  
 _(automatic message)_

PR #705, opened 2023-08-24
Main

chandachewe10:
This PR adds a list of missing currencies from the list of currencies array in `setup.php`. Some currencies for the countries which are available like Zambia (ZMW), Rwanda(RWF) , Malawi(MWK) are not available. People leaving in these countries might find it difficulty to use the software in their own local currency


johnnyq:
Hi @chandachewe10 Thank you for the pull request however you are missing some single quotes around some of the currencies please fix before we can merge.

chandachewe10:
@johnnyq kindly check

johnnyq:
looks good @chandachewe10 reeling on in

PR #706, opened 2023-08-26
Copy Updated Currencies from setup.php to get_settings.php

chandachewe10:
Copy Updated Currencies from `setup.php` to `get_settings.php`.


chandachewe10:
There is duplicated or repetitive code within a software program with respect to constants (countries, currencies and locales).
 Its occurring in `setup.php` and `get_settings.php`. For example there is `$countries_array` list, 
`$currencies_array` and `$locales_array` list in `setup.php` but again these array list are manually typed repeated in 
get_settings.php as we have the same constants:  `$countries_array` list, 
`$currencies_array` and `$locales_array` list.

This is becoming hard to mantain cause when you change a currency for example in `setup.php` you have to change all files (again in `get_settings.php`) 
which is a little bit of work since there are some files using the currencies located in `get_settings.php`. Hence 
for efficiency, readability, and maintainability I think these files can have a separate directory where they can be imported.

johnnyq:
@chandachewe10 Thank you for the pull request and I agree with you, maybe a file called inc_currencies.php one called inc_locales.php

PR #709, opened 2023-08-27
Add notify by email function when a new ticket is created

wrongecho:
Send an email notification (probably to a shared mailbox/distribution list) when a new ticket is created.

This was requested via forum: https://forum.itflow.org/d/222-ticketing-system-notify-by-email-when-new-ticket-is-created

_Note that we seemingly have to strip smileys out of messages to be able to insert content into the mail queue table. The error I was getting was: ```PHP Fatal error:  Uncaught mysqli_sql_exception: Incorrect string value: '\\xF0\\x9F\\x98\\x8A</...' for column `a_itflow`.`email_queue`.`email_content` at row 1```. See functions.php for the code._

Additionally fixed an error in defining `$session_contact_primary` since the schema has changed.

---

image

image

image

image

image


wrongecho:
Test these changes at: https://newticketnotifications709.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
I have ran into issues with the new mail queue similar to smiles and im sure there are others as well

The way I handled it was using using mysqli_escape_string($mysqli, for both subject and body
and all the vars that were injected in the body or subject were unfiltered but POST vars that needed to be filtered for logging would be named var_name_escaped. This  should eliminate the need to filter using the emojis function.

johnnyq:
similar to here
https://github.com/itflow-org/itflow/blob/c070a14b0ecf6624901abd67fb20eb36e5adc79b/post/ticket.php#L241-L343

wrongecho:
I did try using Sanitize Input already without success. I'll have to look into this further.

Let me know if we can merge in the meantime or not, I'm conscious that the database changes (version updates) makes this a little time sensitive / blocking other changes.

johnnyq:
Hi @wrongecho yeah our sanitize function will cause an issue as it strips tags needed then mysqli_escape_string puts the wrong slashes after it is stripped causing breakage of all sorts. A simple mysqli_escape should do.
I'm going to reel this one, will take a further look later on.

wrongecho:
Perfect, thanks! :)

I was just going for MVP to get the feature working 90%, we can always perfect it later.

PR #710, opened 2023-08-27
Forgot Password From Admin

chandachewe10:
Thought this is mandatory and obvious for every login page. If the user has forgotten their password there will be no Login In until manual intervention is done on the database. This functionality will allow the user who has forgotten their password to reset (via email) and create a new one. A new `password_resets` table has been added.

### DIFF
--- a/2023-08-27__backup.sql
+++ b/2023-08-27__backup.sql
@@ -1,0 +1,1409 @@
+
+
+CREATE TABLE `accounts` (
+  `account_id` int(11) NOT NULL AUTO_INCREMENT,
+  `account_name` varchar(200) NOT NULL,
+  `opening_balance` decimal(15,2) NOT NULL DEFAULT 0.00,
+  `account_currency_code` varchar(200) NOT NULL,
+  `account_notes` text DEFAULT NULL,
+  `account_created_at` datetime NOT NULL DEFAULT current_timestamp(),
+  `account_updated_at` datetime DEFAULT NULL ON UPDATE current_timestamp(),
+  `account_archived_at` datetime DEFAULT NULL,
+  PRIMARY KEY (`account_id`)
+) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
+
+INSERT INTO accounts VALUES("1","COMPANY ACCOUNT","0.00","ZMW","This is the Company Bank Account","2023-08-26 12:34:15","2023-08-26 16:47:17","");
+INSERT INTO accounts VALUES("2","Chanda Chewe","0.00","ZMW","This is the personal Account","2023-08-26 16:47:46","","");
+
+
+
+CREATE TABLE `api_keys` (
+  `api_key_id` int(11) NOT NULL AUTO_INCREMENT,
+  `api_key_name` varchar(255) NOT NULL,
+  `api_key_secret` varchar(255) NOT NULL,
+  `api_key_created_at` datetime NOT NULL DEFAULT current_timestamp(),
+  `api_key_expire` date NOT NULL,
+  `api_key_client_id` int(11) NOT NULL DEFAULT 0,
+  PRIMARY KEY (`api_key_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;
+
+
+
+
+CREATE TABLE `asset_custom` (
+  `asset_custom_id` int(11) NOT NULL AUTO_INCREMENT,
+  `asset_custom_field_value` int(11) NOT NULL,
+  `asset_custom_field_id` int(11) NOT NULL,
+  `asset_custom_asset_id` int(11) NOT NULL,
+  PRIMARY KEY (`asset_custom_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;
+
+
+
+
+CREATE TABLE `asset_documents` (
+  `asset_id` int(11) NOT NULL,
+  `document_id` int(11) NOT NULL,
+  PRIMARY KEY (`asset_id`,`document_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;
+
+
+
+
+CREATE TABLE `asset_files` (
+  `asset_id` int(11) NOT NULL,
+  `file_id` int(11) NOT NULL,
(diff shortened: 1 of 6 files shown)


wrongecho:
I hate to be the bearer of bad news, as its clear you've put lots of work into this but how are you handling asset logins/password encryption?

After a user resets their password, they won't be able to access the encrypted login entries (usernames/passwords)?

See https://forum.itflow.org/d/76-forget-password-reset-via-email-link/3

johnnyq:
@chandachewe10 @wrongecho is right. This would have be implemented somehow into the code but the problem is you have to be able to login to decrypt the master password to decrypt the client passwords that are stored. This is where this will get tricky.

chandachewe10:
I forgot I raised a similar thought sometime back, its been long here. So what could happen if the decryption key is compromised?
Does it means all encrypted passwords would be vulnerable? 
and what if the available administrators having access to the decryption key also
forgets their passwords? Does it mean they will not log in? Even so password resets might become challenging if there are no available administrators.
What's wrong with the normal bcrypt hashing without the further encrypting them ?

wrongecho:
We can't just hash the usernames/passwords stored in logins as hashing is a one-way function. We need encryption, and encryption comes with the challenge of key management.

To compromise any logins stored in ITFlow, you'd need both a dump of the database (or at least the logins table) **AND** the master encryption key. 
The master encryption key isn't stored anywhere in plaintext, but protected by each users password as previous discussed. The downside to this is that any operations involving the stored logins or the users password must be done by a user with access to the master decryption key. The alternative is storing the master decryption key somewhere on disk or in the database - this is the best option I could see without requiring a HSM/Third party service.

If all administrators forget their passwords and the decryption key wasn't backed up then any encrypted logins are lost. This is entirely by design. Having to spend a few hours resetting passwords at your (relative) convenience is nothing compared to having them all dumped out in plaintext and scrambling to do damage control. It's my opinion that architecting ITFlow secret/login storage any other way would be reckless, but we are of course always open to discussion as this does add a few limitations.

Please see https://github.com/itflow-org/itflow/issues/264 for more info.

chandachewe10:
I really don't understand why you are hashing the clients Login Entries with the Master Key + salting which to me seems like re-inventing the wheel (salting is already done by BCRYPT by default).

Under normal circumstances a password shouldn't be reversed and  Bcrypt is a one-way function, which means that it's not possible to reverse-engineer the original password from the hash. This is essential for securely storing user passwords, as even the system administrators shouldn't be able to recover users' original passwords.

wrongecho:
> I really don't understand why you are hashing the clients Login Entries with the Master Key + salting which to me seems like re-inventing the wheel (salting is already done by BCRYPT by default).

We're not hashing, we're encrypting. The client username/password login entry is encrypted using the master key. 
The master key is encrypted using the hashed password of the ITFlow agent and is then stored in the database as the user specific key.

> Under normal circumstances a password shouldn't be reversed  

I agree, but we want to actually be able to see/use the stored client login usernames and passwords in ITFlow. Hashing requires you to know & match the input which kinda defeats the purpose of documenting them at all?
The password the agent uses to login to ITFlow cannot be recovered by anyone.

chandachewe10:
> The alternative is storing the master decryption key somewhere on disk or in the database - this is the best option I could see without requiring a HSM/Third party service.

The alternative is storing the master key somewhere  (if possible) - this is the best option rather than losing access to the system because all admins have gotten their password(s). This way when an admins password is reseted, the new password can be used to encrypt the master key as a user cipher text as usual.

chandachewe10:
The alternative is storing the master key somewhere (if possible) - this is the best option rather than losing access to the system because all admins have gotten their password(s). This way when an admins password is reseted, the new password can be used to encrypt the master key as a user cipher text as usual.

johnnyq:
@chandachewe10 We found that to be insecure as well as a server could be compromised and key stolen. I think a good recommendation would be to create another admin use with a long password and store the creds somewhere safe. Its security over usability when we can.

chandachewe10:
> @chandachewe10 We found that to be insecure, as a server could be compromised and key stolen. I think a good recommendation would be to create another admin user with a long password and store the creds somewhere safe. Its security over usability when we can.

@johnnyq storing credentials outside the system is even worse as someone might peep on them.  I hope a seamless work around will be found, your process can work well in large organizations with a lot of administrators but where there is one or two admins its very risky, Imagine we have encrypted 100 usernames/passwords of clients software's and both of us (only 2 admins) forgot our passwords. We are doomed.  Am closing this for now

wrongecho:
Backing up your database and master encryption key is the best way to prevent the scenario you mention.

chandachewe10:
> You should back up the master key/password soon after install - it never changes.
> 
> You then take periodic database backups
> 
> Even if you've forgotten your login details, you can still decrypt the logins stored in ITFlow, as you have the database and master key.

Unless I backup  the clear text of my password. Assuming I lose my login credentials today (email and password) then the db will probably contain my hashed password (BCRYPT) and my user_specific_encryption_key (encrypted MASTER KEY). Am assuming this is my backed up db. Now at this point I have already encrypted my clients assets passwords and usernames using that master key. To retrieve those assets logins of my clients I need my master key again (which I should get using my **unhashed clear text password**). Note that at this point have gotten my clear text plain password. How is the hashed bcrypt password from the backed up db going to assist me to get the master key from the user_cipher_text?

wrongecho:
Please see your previous forum post and the original Github issue/PR/code for an in depth understanding of how the encryption works. I've gone over it a few times previously :) 

https://forum.itflow.org/d/76-forget-password-reset-via-email-link/3

#264


If you ever do actually need to decrypt database entries using the master key, there is a script I made a while back: https://github.com/itflow-org/itflow-misc/blob/main/ITFlow%20Password%20Decryption%20Example.php

wrongecho:
> It doesn't explain how you can retrieve the original master key from the backup database in a case where yo don't have the plain text password

You don't, that's the entire point. This is why you back up your master encryption key somewhere safe.

chandachewe10:
> > It doesn't explain how you can retrieve the original master key from the backup database in a case where yo don't have the plain text password
> 
> You don't, that's the entire point. This is why you back up your master encryption key somewhere safe.

Which to me sounds risk of course! You can't trust a user to keep it somewhere safe.

wrongecho:
ITFlow's primary audience is system administrators, not users.

wrongecho:
Unfortunately nothing in life is without risk. 

We are open to ideas on how to improve the master key management in a secure way, but until then I'd recommend you print out a few copies of your master encryption key and store them in secure, fireproof safes.

PR #713, opened 2023-09-03
Auditing bugfix

wrongecho:
Ticket creation logging is now associated with the client that the ticket was raised under


wrongecho:
Test these changes at: https://anonymisecontact2713.pr-review.itflow.org  
 _(automatic message)_

PR #717, opened 2023-09-11
Updater: Show previously applied updates (git commits)

wrongecho:


wrongecho:
Needs some more testing before we commit, breaking the updater would be annoying!

wrongecho:
Test these changes at: https://updatershowprevious717.pr-review.itflow.org  
 _(automatic message)_

wrongecho:
Won't have time to test this thoroughly enough and don't want to break the updater before 1.0. Code can remain here for future reference if we do ever implement.

PR #725, opened 2023-09-30
Fix client delete function

wrongecho:
Bugfix + Enhancement

When testing the recent changes, I noticed the restored client delete modal wasn't properly directing people to the `post.php?delete_client` page - it would just hang after you click Yes.

Whilst working on fixing this using the new confirmation modals, I figured that deleting an **entire clients data** shouldn't be as easy as clicking "Yes", so restored the old confirmation dialog where you're required to enter the client name (complete with CSRF). 
Hope you agree that this is the best way forward for this specific extremely destructive function! :)

image

image

### DIFF
--- a/client_delete_modal.php
+++ b/client_delete_modal.php
@@ -5,7 +5,7 @@
                 <div class="mb-4" style="text-align: center;">
                     <i class="far fa-10x fa-times-circle text-danger mb-3 mt-3"></i>
                     <h2>Are you really, really, really sure?</h2>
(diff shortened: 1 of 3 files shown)


wrongecho:
Test these changes at: https://fixclientdelete725.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
This looks fantastic!! =] I like that you have to enter the client name in to continue

PR #727, opened 2023-09-30
Users - Prevent modifying yourself via users.php, should use profile

wrongecho:
This PR prevents users from accidentally disabling/archiving themselves via the admin settings page.

Self name/email/password changes should be done via profile settings.


wrongecho:
Test these changes at: https://settingsuserpreventselfmod727.pr-review.itflow.org  
 _(automatic message)_

PR #729, opened 2023-10-01
Client logins/password - Add tracking in DB when passwords are changed

wrongecho:
Currently, we only record when login entries are accessed or updated, but a login entry being "updated" does not necessarily mean the password was changed (e.g. the description could have just been updated).

This PR adds an additional database field that allows you to track exactly when a login entry password was last changed.

In the future, I'd like to create a technical report that allows you to see what passwords haven't been rotated in X days, as well as exposing the last change date somewhere on the client_login.php page


wrongecho:
Test these changes at: https://loginpasswordchangedat729.pr-review.itflow.org  
 _(automatic message)_

PR #730, opened 2023-10-01
Client logins/passwords - Ajax change + logging

wrongecho:
A few changes to the credential manager TOTP function:

- The "hover to view" function now sends the login_id to ajax.php, rather than the TOTP secret
- Viewing the TOTP code is now audited in the logs under Login/View TOTP


wrongecho:
Test these changes at: https://improvetotpsecurity730.pr-review.itflow.org  
 _(automatic message)_

PR #731, opened 2023-10-02
Client logins/passwords - Add basic password rotation report

wrongecho:
This basic report shows you which client login entries have not had their passwords changed/rotated in the last 90 days. 

Password rotation is no longer encouraged for users memorising their _own_ passwords. However, password rotation is essential for service/shared accounts commonly used by MSPs in situations where individual accounts aren't available/viable.

### DIFF
--- a/report_password_rotation.php
+++ b/report_password_rotation.php
@@ -1,0 +1,73 @@
+<?php
+
+require_once("inc_all_reports.php");
+validateTechRole();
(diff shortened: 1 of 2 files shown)


wrongecho:
Test these changes at: https://passwordrotationreport731.pr-review.itflow.org  
 _(automatic message)_

PR #732, opened 2023-10-07
Email parsing for all domains registered under a client

wrongecho:
- **Add support for email parsing/contact creation for all domains registered under a client in the domains module, rather than just the client main website.**

- Additionally fix domain_created_at bug and move the new ticket auto-reply message to the email queue instead

_Future work: Make ticket parsing work with HTML emails (HTML emails break agent notifs)_


wrongecho:
Test these changes at: https://emailparsedomains732.pr-review.itflow.org  
 _(automatic message)_

PR #733, opened 2023-10-07
Ticket updates - Default to internal update if contact is agent

wrongecho:
When adding a ticket reply, default the reply to internal (and don't show the box to add a public reply) if the contact email isn't set **or** if the contact email matches the currently logged in agent.

https://forum.itflow.org/d/420-ticket-notifications/4


wrongecho:
Test these changes at: https://ticketupdatetypeemailmatch733.pr-review.itflow.org  
 _(automatic message)_

PR #736, opened 2023-10-07
Add function to reset all user/agent passwords in case of IR

wrongecho:
In the case of a potential security incident, this PR adds the ability to force all agent passwords are changed and distributed manually by the system administrator. 

image

_You must have more than 1 ITFlow user for the option to show on Settings > Users. The modal could still use some CSS love for the password box_

### DIFF
--- a/post/user.php
+++ b/post/user.php
@@ -280,3 +280,47 @@ if (isset($_POST['export_users_csv'])) {
     exit;
 
 }
+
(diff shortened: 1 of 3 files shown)


wrongecho:
Test these changes at: https://iragentforcepwreset736.pr-review.itflow.org  
 _(automatic message)_

PR #747, opened 2023-10-10
Update technical dashboard links.

o-psi:
Made it so that the Admins can access the technical dashboard from sidebar, rather than just the financial dashboard. 
Our president is not very hands on with accounting, but is with operations.


wrongecho:
Test these changes at: https://technicaldashboardforadmins747.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
makes sense merging in

PR #748, opened 2023-10-10
Update guest quote view to make more sense

o-psi:
Moved accept decline buttons to bottom of page, below footer. Changed logic to disallow approvals and denials on draft status.


wrongecho:
Test these changes at: https://quoteguestview748.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Yes this makes sense too merging

PR #752, opened 2023-10-12
[Urgent] Fix Foreign Key for account types

o-psi:
Noticed some weird behavior on new installs, looks like the constraint was messed up from order.

apache log was full of 
`[Thu Oct 12 15:04:05.258672 2023] [php:error] [pid 30423] [client 136.50.97.27:63652] PHP Fatal error:  Uncaught mysqli_sql_exception: Cannot add or update a child row: a foreign key constraint fails (`itflow`.`accounts`, CONSTRAINT `fk_accounts_account_type_id` FOREIGN KEY (`account_type`) REFERENCES `account_types` (`account_type_id`)) in /var/www/..x../setup.php:1022\nStack trace:\n#0 /var/www/..x../setup.php(1022): mysqli_query()\n#1 {main}\n  thrown in /var/www/..x../setup.php on line 1022, referer: https:/..x../setup.php?company`

Moved account_types above the accounts table in the DB.sql declaration.

### DIFF
--- a/db.sql
+++ b/db.sql
@@ -19,6 +19,19 @@
 -- Table structure for table `accounts`
 --
 
+DROP TABLE IF EXISTS `account_types`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `account_types` (
+  `account_type_id` int(11) NOT NULL AUTO_INCREMENT,
+  `account_type_name` varchar(200) NOT NULL,
+  `account_type_description` text DEFAULT NULL,
+  `account_type_created_at` datetime NOT NULL DEFAULT current_timestamp(),
+  `account_type_updated_at` datetime DEFAULT NULL ON UPDATE current_timestamp(),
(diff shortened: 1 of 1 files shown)


wrongecho:
Test these changes at: https://fixaccounttypefk752.pr-review.itflow.org  
 _(automatic message)_

o-psi:
@wrongecho @johnnyq this should be merged before the next person installs...

johnnyq:
Im wondering if this is going to be an issue down the road when we do mysql structure dump via the command line as we update the db further. 
it will just place accounts before account_types again. We never actually used the Foreign key option before.
So typically after we make DB changes we execute:
`mysqldump -u itflow_dev -p -h localhost --opt itflow_dev -d --single-transaction | sed 's/ AUTO_INCREMENT=[0-9]*\b//' > db.sql`
If you dont need the FK we should drop it for sure to prevent breakage

o-psi:
FK dropped in next PR, the dropdowns are resilient enough to handle unknowns that could be caused without it.

PR #755, opened 2023-10-12
Service Templates

o-psi:
This PR adds all the groundwork for service templates, additionally, lets you assign a price to the service(no function yet)

### DIFF
--- a/client_service_add_modal.php
+++ b/client_service_add_modal.php
@@ -1,232 +1,281 @@
 <div class="modal" id="addServiceModal" tabindex="-1">
-  <div class="modal-dialog modal-md">
-    <div class="modal-content bg-dark">
-      <div class="modal-header">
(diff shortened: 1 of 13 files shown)


wrongecho:
Test these changes at: https://servicetemplates755.pr-review.itflow.org  
 _(automatic message)_

o-psi:
This looks like a lot of code, but much is just changing spacing to follow standards in documents I am in.

johnnyq:
Hello @o-psi Instead of creating a whole new table since most of the fields are the same as a service template
Add a tinyint (1) field to services and call it service_template: 1 meaning its a template 0 meaning its not.
We did the same with document templates, vendor templates and software licenses templates. This can be replicated here as well.

wrongecho:
Can I suggest we don't include anything about pricing/service costs until it's actually decided how that will work and is fully functional? It just seems to confuse users and we get "bug" posts.

PR #758, opened 2023-10-13
Add archive option for account types

o-psi:
### Account Type Archiving:
Added a check to detect if archive_account_type is set in the $_GET request.
When set, the respective account type is marked as archived by updating the account_type_archived_at column in the account_types table with the current timestamp.
**Logging:**
When an account type is archived, a new entry is added to the logs table.
Type: Account Type
Action: Archive
Description: Contains the archived account's ID.
IP & User Agent: Captured from the session data.
**Notification & Redirection:**
A session alert message "Account Archived" is set after successful archiving.
User is then redirected back to the referring page.

### DIFF
--- a/post/account_type.php
+++ b/post/account_type.php
@@ -48,7 +48,7 @@ if (isset($_POST['add_account_type'])) {
 
 
     //Logging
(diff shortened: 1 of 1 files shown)


wrongecho:
Test these changes at: https://addarchiveforaccounttypes758.pr-review.itflow.org  
 _(automatic message)_

PR #759, opened 2023-10-13
Assets report

o-psi:
This PR introduces a feature to display all assets grouped by clients. Users can also print this list directly from the interface. The key components of this change are:

**Database Queries:**

A query to fetch all clients sorted by their names.
For each client, a subsequent query retrieves their associated assets, sorted by asset name.

**User Interface:**

Assets are displayed in a table format within a card component.
Each asset entry provides a direct link to view all assets for a specific client and a link to view the details of a particular asset.
Users have the option to print the list via a button in the card's header.


wrongecho:
Test these changes at: https://AssetsReport759.pr-review.itflow.org  
 _(automatic message)_

PR #762, opened 2023-10-14
Sort invoice items

o-psi:
This PR adds the ability to order invoice items.

I know we wanted to do draggable, but I need to practice JS more for that.

### DIFF
--- a/database_updates.php
+++ b/database_updates.php
@@ -1407,19 +1407,19 @@ if (LATEST_DATABASE_VERSION > CURRENT_DATABASE_VERSION) {
     //
     if (CURRENT_DATABASE_VERSION == '0.8.8') {
     // Insert queries here required to update to DB version 0.8.9
-    mysqli_query($mysqli, "ALTER TABLE `invoice_items` ADD `item_order_id` INT(11) NOT NULL DEFAULT 0 AFTER `item_total`");
-    // Update existing invoices so that item_order_id is set to item_id
+    mysqli_query($mysqli, "ALTER TABLE `invoice_items` ADD `item_order` INT(11) NOT NULL DEFAULT 0 AFTER `item_total`");
+    // Update existing invoices so that item_order is set to item_id
     $sql_invoices = mysqli_query($mysqli, "SELECT invoice_id FROM invoices WHERE invoice_id IS NOT NULL");
     foreach ($sql_invoices as $row) {
         $invoice_id = $row['invoice_id'];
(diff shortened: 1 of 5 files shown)


wrongecho:
Test these changes at: https://Invoiceitemsorder762.pr-review.itflow.org  
 _(automatic message)_

database_updates.php
```diff
@@ -1405,13 +1405,29 @@
...
+    mysqli_query($mysqli, "ALTER TABLE `invoice_items` ADD `item_order_id` INT(11) NOT NULL DEFAULT 0 AFTER `item_total`");
+    // Update existing invoices so that item_order_id is set to item_id
+    $sql_invoices = mysqli_query($mysqli, "SELECT invoice_id FROM invoices WHERE invoice_id IS NOT NULL");
+    foreach ($sql_invoices as $row) {
+        $invoice_id = $row['invoice_id'];
+        $sql_invoice_items = mysqli_query($mysqli, "SELECT item_id FROM invoice_items WHERE item_invoice_id = '$invoice_id' ORDER BY item_id ASC");
+        $item_order_id = 1;
+        foreach ($sql_invoice_items as $row) {
+            $item_id = $row['item_id'];
+            mysqli_query($mysqli, "UPDATE invoice_items SET item_order_id = '$item_order_id' WHERE item_id = '$item_id'");
```
o-psi, line 1419: Update each invoice item and give it a, item order id

johnnyq:
Nice work, you dont slow down hahhaa but can we rename item_order_id to just item_order we use id to reference items in multi tables and it would cause a little bit of confusion

o-psi:
@johnnyq Done :)

johnnyq (approved):

johnnyq (approved):

johnnyq:
So far so good on the item order I pulled it in it looks fantastic but needs some more functionality love
Item order should not be greater than the amount of items in an invoice as I can keep clicking the up and down arrows as many times to increment the order.

Some Logic to add

- If order is 0 do not show Up Arrow as this would be the first item
- If order is the last order do not show down arrow.

Of course once this is super stable we should replicate this everywhere recurring invoices, quotes, quote to invoice, cron generated invoices from recurring etc.

o-psi:
@johnnyq See https://github.com/itflow-org/itflow/pull/768
 
 I believe this fixes your stability concern as it reduces complexity substantially, and additionally removes the arrows as suggested.

PR #766, opened 2023-10-14
Allow Unarchiving and Deleting Categories.

o-psi:
This PR allows for deleting and unarchiving archived categories.

### DIFF
--- a/categories.php
+++ b/categories.php
@@ -15,28 +15,20 @@ if (isset($_GET['category'])) {
 //Rebuild URL
 $url_query_strings_sort = http_build_query($get_copy);
 
-if (isset($_GET['archived'])) {
-    $sql = mysqli_query(
-        $mysqli,
-        "SELECT SQL_CALC_FOUND_ROWS * FROM categories
-        WHERE category_name LIKE '%$q%'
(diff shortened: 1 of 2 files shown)


wrongecho:
Test these changes at: https://categoriesarchived766.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Hello @o-psi We already created dudup code logic to display archived and un-archived data.

In pagination_head.php which is used in almost all the table views already takes care of the archived bit see here
// Archived
```

$archived = 0;

if (isset($_GET['archived'])) {
    $archived = intval($_GET['archived']);
}

if($archived == 1){
    $archive_query = "archived_at IS NOT NULL";
} else {
    $archive_query = "archived_at IS NULL";
}
```

This is currently being used in client_contacts.php so you can study how the logic works there. most of the logic is at the top.

Just didn't want you reinventing the wheel ;-)
But keep up the great work we sure can use the unarchive, archive post data for sure.
Please update to your code to take a advantage of the function we already have.

o-psi:
Love that that code is deduped, and yall thought of a better way to do it than I. This has been fixed @johnnyq

PR #772, opened 2023-10-17
Leads

o-psi:
This pull request (PR) introduces a comprehensive update to our client management system by incorporating a "Leads" feature. This new feature involves several key changes and enhancements outlined below:

Database Structure Enhancement: The clients' database has been expanded to include a "Leads" field, allowing for more robust data collection and management.

User Interface Updates for Leads Management: Modifications have been made to the system's user interface (UI). These changes involve the add and edit modal dialogs, which have been updated to handle leads information, providing a more integrated experience for users when managing client data.

Introduction of Leads Management Screen: A new management screen dedicated to leads has been introduced. This screen provides a detailed view and management options specifically for leads, enhancing usability and productivity.

Refinement of Client Listing: The client listing functionality has been refined to allow for the filtering out of leads. This update ensures a more streamlined view, enabling users to differentiate between regular clients and leads.

Leads Integration in Sales and Client Model: The new leads feature has been fully integrated into the existing sales processes and client models. This ensures that leads are part of the holistic client management experience, allowing for more seamless interaction and data consistency.

Backend Updates for Leads Support: Necessary backend updates have been made, particularly to the POST endpoints, to support the addition and management of leads data. This ensures the backend can adequately handle and process the new data entries related to leads.

### DIFF
--- a/client_edit_modal.php
+++ b/client_edit_modal.php
@@ -2,7 +2,9 @@
     <div class="modal-dialog">
         <div class="modal-content bg-dark">
             <div class="modal-header">
-                <h5 class="modal-title"><i class="fa fa-fw fa-user-edit mr-2"></i>Editing: <strong><?php echo $client_name; ?></strong></h5>
+                <h5 class="modal-title"><i class="fa fa-fw fa-user-edit mr-2"></i>Editing: <strong>
+                        <?php echo $client_name; ?>
+                    </strong></h5>
                 <button type="button" class="close text-white" data-dismiss="modal">
                     <span>&times;</span>
                 </button>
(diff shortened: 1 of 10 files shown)


wrongecho:
Test these changes at: https://leads772.pr-review.itflow.org  
 _(automatic message)_

client_edit_modal.php
```diff
@@ -81,10 +100,26 @@
                                     <div class="input-group-prepend">
                                         <span class="input-group-text"><i class="fa fa-fw fa-globe"></i></span>
                                     </div>
-                                    <input type="text" class="form-control" name="website" placeholder="ex. google.com" value="<?php echo $client_website; ?>">
+                                    <input type="text" class="form-control" name="website" placeholder="ex. google.com"
+                                        value="<?php echo $client_website; ?>">
+                                </div>
+                            </div>
+
+                            <div class="form-group">
```
o-psi, line 108: I wanted to use the following switch from AdminLTE, but could never figure out how to get it to show as checked.
if someone else wants to take a crack at it...

```
                                <div class="input-group">
                                    <div class="custom-control custom-switch">
                                        <input class="custom-control-input" id="customSwitch1" type="checkbox" name="lead" value="1"<?php if ($client_is_lead == 1) {
                                            echo " checked";
                                        } ?>><label class="custom-control-label" for="customSwitch1">Is Lead?</label>
                                    </div>
                                </div>
```

wrongecho:
Interesting ideas here!

Whilst I appreciate your time, end to end sales pipeline management wasn't in my vision for ITFlow's future. Not too sure about Johnny. 

This may be something worth raising on the forum. 

If we _were_ to implement, I'm not too sure on shoehorning this into the existing clients functionality - e.g. How do I note a call with a lead against their contact? How do you convert a lead to a prospect? How do you create a deal? How do you mark a deal as won/lost? 
There are a lot of extra required features and the overarching clients module isn't the right place for them - in my view.

o-psi:
If we are doing an end to end sales pipeline management, then I agree clients is not the right place for it. But that's not what this PR is for. 

I think a simple differentiation between regular clients and leads will be great for our target audience. I think a full sales pipeline is too complex unless you have a sales team. This way we can separate out any of the companies that internally we do not refer to as clients. 

See forum post: https://forum.itflow.org/d/472-lead-management-update

o-psi:
@johnnyq What are your thoughts?

johnnyq:
You know @o-psi  I like it. I have clients I added that never became a client just a lead so this will cleanup the mess of non clients.

o-psi:
@wrongecho @johnnyq The more I think about it, the better it is that leads are just clients. We like to offer free estimates to lure sales. Therefore, if we can print the documentation report for the lead before we score the client, its a big sale/very powerful.
Basically it allows for showing the leads what we can do for them.

wrongecho:
I don't disagree, it's just the wording of "Lead" would make people think more Salesforce. That was where I thought you wanted to go with this. 

Given that this is to differentiate between groyps of clients and not for end to end sales pipeline perhaps the term "Pre onboarded" or something like that is clearer? 

I was thinking we can do the same for clients that have left and are inactive. Clients page would default to your active clients but your Potential new clients and old clients are only a click away. 

https://forum.itflow.org/d/249-new-feature-request-activeinactive-client-checkbox

o-psi:
@wrongecho I adore the inactive idea. I also am all for renaming leads, and while pre-onboarded is descriptive, it doesn't quite fit in with the rest of the names.

PR #776, opened 2023-10-19
Discounts

o-psi:
Added discounts to the invoices. This was a crucial feature for us, as we give discounts to clients for referrals.


wrongecho:
Test these changes at: https://discounts776.pr-review.itflow.org  
 _(automatic message)_

wrongecho:
This has been on the to do list for a while. Just looking over the files changes, I'm not seeing any references to Stripe or Cron being updated?

johnnyq:
Nice work on the discounts but as @wrongecho stated it is in-complete and I fear there may be some unpredictable actions that could occur and break some of the functions
We handle discounts very differently but it does work
we will create a Line item labeled Discount. For the Description we state why were discounting with Quantity of 1 and price be a negative number which will subtract off their bill. 
I think the addition of this code will add to complexity of it.
We are open to your feedback on this @o-psi

johnnyq:
see here for an example
image

o-psi:
Make sure to have your accountant caps on for this one! Haha. Discounts are removed from the subtotal, not the total. 

Since these discounts are added and calculated whenever an item is added, or when the discount is applied; the total of the invoice itself (in the db) is being updated.

That line item in the database will mess up any line item reports I plan on adding later. That is the reason that in most QuickBooks classes, they tell you to never do what you are recommending.

@wrongecho From what I saw in the codebase, payments are only calculated from the balance. See below screenshot of stripe using the correct amount to charge.

o-psi:
image

o-psi:
> Nice work on the discounts but as @wrongecho stated it is in-complete and I fear there may be some unpredictable actions that could occur and break some of the functions
> We are open to your feedback on this @o-psi 


Can I ask what led you to believe it is incomplete? Or what you are concerned will break?
This has been in development for about 4 days, had a breakdown of needs, and a related scrum board.
We applied this PR to our production environment this morning with no adverse consequences. 

Seems a bit fast to say it's not done, before you have done any testing. I had to update the DB on the demo  after those comments were made. :/

johnnyq:
Hi @o-psi You bring up a good point regarding the accounting hats and future reports, along with other invoice software clearly implementing discounts. Which totally makes sense.
The reason why I brought up that this was incomplete were for reasons I listed below:
- Discounts isn't implemented in Export Invoice PDF
- Quotes isn't fully implemented with discounts along with the logic Quote to Invoice
- Recurring isnt fully implemented with it take the recurring discount and applying it to the next recurring invoice when Cron executes.
- I Didn't realize you tested it so thoroughly in-house

But after thinking more into it, this was your intention as I believe you wanted to start with invoices then work into the quotes and recurring, like you did with sorting.

You are also very extremely fast to respond when things do go south.

Again my apologies I didn't mean no disrespect, your doing great.

Ill do some more testing reel this PR in.

wrongecho:
> image
> This is on the demo for this PR. Give it a shot here: [PR Review Link](https://discounts776.pr-review.itflow.org/invoice.php?invoice_id=2)

Ah I see. It's been a while since I wrote that and I'm limited to testing on mobile tonight. 

Can we please make it clear on the left hand side of the Stripe pay that a discount is applied? It only shows the invoice items.

johnnyq:
other than making the discount clear it looks good to me,

I just reeled in Leads so that database_version.php will need to increment to .1 and database update will need to go up 1 more as well.

o-psi:
> * Discounts isn't implemented in Export Invoice PDF
The code I have for this is not good enough for implementation, have not been able to get the formatting correct.

> * Quotes isn't fully implemented with discounts along with the logic Quote to Invoice
> * Recurring isnt fully implemented with it take the recurring discount and applying it to the next recurring invoice when Cron executes.
> But after thinking more into it, this was your intention as I believe you wanted to start with invoices then work into the quotes and recurring, like you did with sorting.
This is exactly the case. Any comments that we decide on for invoices can be applied to the others.

> You are also very extremely fast to respond when things do go south.
I do try. Even though this is GPL and all, I recognize the amount of work both of y'all (@wrongecho and @johnnyq) have put in. This is your baby, and I promise on my mommas grave that I will treat it like such. Excuse my southern expressions

> Again my apologies I didn't mean no disrespect, your doing great.
I took no disrespect, and greatly appreciate your response and attitude in such. 

> Ill do some more testing reel this PR in.
Please do. I think this puts ITFlow more on par with something like Invoice Ninja (the tool we are moving from)

o-psi:
> @wrongecho 
 Can we please make it clear on the left hand side of the Stripe pay that a discount is applied? It only shows the invoice items.
> @johnnyq 
>  other than making the discount clear it looks good to me,

This has been done in the same manner as the "Already Paid"

> I just reeled in Leads so that database_version.php will need to increment to .1 and database update will need to go up 1 more as well.

This has also been updated

wrongecho:
Looks like the DB change broke the PR Review again! But just looking at the code change I think that should do it. 

> I had to update the DB on the demo, after those comments were made. :/

Not sure what you meant by this. 
For the **demo**: There's no automation to update the demo, it's purely there as a "I want to try ITFlow quickly". 
Please don't take the demo not being updated with your changes personally. I only ever update it when I want someone to try and reproduce a bug on it.


My earlier comments around Cron/Stripe were just out of an abundance of caution for things to check that might have been missed, as other recent PRs have broken the installer in unforseen weird and wonderful ways. Such is life.

PR #781, opened 2023-10-21
Add background ajax query in an attempt to prevent session timeout

wrongecho:
Whilst this isn't something I've run into myself, we've had a few posts on the forum about sessions timing out when working on things like editing documents.

This PR adds a AJAX background query to keepalive.php every 10 mins which _should_ keep sessions alive.


wrongecho:
Test these changes at: https://sessiontimeoutgarbageworkaround781.pr-review.itflow.org  
 _(automatic message)_

PR #783, opened 2023-10-21
Expand range of allowed file extensions for email to ticket attachments

wrongecho:
These match those suggested in the client_file_upload_modal


wrongecho:
Test these changes at: https://emailtoticketextensions783.pr-review.itflow.org  
 _(automatic message)_

PR #785, opened 2023-10-24
Update Timer to stop when in focus

o-psi:
Update the JS to make the timer stop when the focus is on it.


wrongecho:
Test these changes at: https://UpdateTimer785.pr-review.itflow.org  
 _(automatic message)_

PR #791, opened 2023-10-25
Save ticket timer to local storage

o-psi:
Stores the ticket timer to local storage to make it so that if the ticket is closed, the time is saved.


wrongecho:
Test these changes at: https://savetimertolocal791.pr-review.itflow.org  
 _(automatic message)_

wrongecho:
This is very neat!

Please ensure to add functionality to clear/reset the timer storage when you submit the ticket, otherwise it just continues the clock with every update.

o-psi:
Good Catch @wrongecho.

@johnnyq any concerns?

johnnyq:
it looks pretty good, although im not 100% familiar with JS either but lets reel it in as I dont see any major issues.

PR #795, opened 2023-11-03
Update global search to include client domains

nellistc:
Add domains to global search. Provide domain and domain expiry date in the results. Link to relevant client domains page.

### DIFF
--- a/global_search.php
+++ b/global_search.php
@@ -261,6 +261,7 @@ if (isset($_GET['query'])) {
                             <tr>
                                 <th>Name</th>
                                 <th>Expiry</th>
+                                <th>>Client</th>
                             </tr>
                             </thead>
                             <tbody>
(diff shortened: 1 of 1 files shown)


wrongecho:
Test these changes at: https://globalsearchdomains795.pr-review.itflow.org  
 _(automatic message)_

wrongecho:
Test these changes at: https://globalsearchdomains795.pr-review.itflow.org  
 _(automatic message)_

nellistc:
Thinking about it now, the client should probably be in the search results - if the same domain is listed again multiple clients.

johnnyq:
> Thinking about it now, the client should probably be in the search results - if the same domain is listed against multiple clients. Is that possible?

Typically we do as the domain belongs to a client. So your right a client field should be present.

nellistc:
> Typically we do as the domain belongs to a client. So your right a client field should be present.

Added the relevant client to the results with a link.

johnnyq:
Looks good were gonna merge this one in, thank you @nellistc

PR #798, opened 2023-11-05
Fix undefined client_is_lead errors

wrongecho:
- Defines client_is_lead within the general client include file to prevent the following error filling up logs on almost every page load.
> "PHP Notice: Undefined variable: client_is_lead in itflow/client_edit_modal.php on line 111"

image


wrongecho:
Test these changes at: https://clientisleadfix798.pr-review.itflow.org  
 _(automatic message)_

PR #799, opened 2023-11-06
footer.php: reset values in all modals after they are hidden

nellistc:
If a modal is hidden or dismissed (cancelled), any modified (but unsaved) values persist when the modal is opened again with the same record.

This change resets the modal when it is hidden so that saved values are shown the next time the modal is opened.

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -12,7 +12,8 @@
 <!-- reset values in modals after hiding/dismissal -->
 <script>
 $(document).ready(function() {
-    $('body').on('hidden.bs.modal', '.modal', function () {
+    // exclude modals having an ID starting with 'add'
+    $('body').on('hidden.bs.modal', 'div.modal:not([id^="add"])', function () {
        // reset the form
        $(this).find('form').trigger('reset');
     });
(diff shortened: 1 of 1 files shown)


wrongecho:
Test these changes at: https://resetmodals799.pr-review.itflow.org  
 _(automatic message)_

wrongecho:
Hey, thanks for the PR!

My only issue with this is that (at least on mobile) modals can be dismissed by clicking outside them. 
This presents an issue when someone is filling in a new/create modal as they then lose all they work they've done.

nellistc:
Yes, I was thinking about that as well. Setting `data-backdrop="static"` stops you from accidentially clicking out of it - maybe an option.

o-psi:
I prefer the modal stays filled. I am working on java script code to save any forms to sessions, so if it the modal has a new ID, it will have the the same, otherwise not.

nellistc:
The latest change leaves the values in for any modals with an ID starting with add. Any others, it resets the form. For me, if I change something in an edit modal and then hit cancel, I don't expect to have the changed value there if I go back in. Worth making it an option in the config?

johnnyq:
closing pull request as we decided against this at this time but we do thank you for your contribution to ITFLow

PR #802, opened 2023-11-12
Time tracking paused when pending

o-psi:
Based on usage, we found the timer doesnt need to stay running when the ticket is marked as pending.


wrongecho:
Test these changes at: https://updatetimerwhenpending802.pr-review.itflow.org  
 _(automatic message)_

PR #804, opened 2023-11-13
Update to allow x frequency instead of assuming 1x

o-psi:
Update to allow x frequency number for recurring invoices instead of assuming 1x.

For example, every 6 months or every 3 years...


wrongecho:
Test these changes at: https://Allowmorefrequencies804.pr-review.itflow.org  
 _(automatic message)_

o-psi:
@johnnyq this should be ready to pull

johnnyq:
Still need to review this one but will merge once reviewed

o-psi:
@wrongecho if you have time, always open to feedback

wrongecho:
I'd need to set some time aside to test the functionality but on initial review the database schema update logic looks incorrect _(breaking recurring invoices entirely)_:
- Wrong version number in `database_updates.php` line [1505](https://github.com/itflow-org/itflow/pull/804/files#diff-63bafdbd7f2d3be9bc97d01f75ca18dc9dce6c08236689026f9ee9d125d2333cR1505)
- No change in `database_version.php` to trigger the update to latest database version

o-psi:
@wrongecho Yea, you're right. I dont know why I struggle with this method of updating the DB. Maybe for 1.0 we can do an array and a function that iterates through the array to apply the needed changes? like "1.0.2" - "UPDATE * ..."

@johnnyq This should be ready.

johnnyq:
Im reviewing this and thinking to myself, this isn't typical for business to have recurring charges daily weekly or every 6 monthes etc.This is very niche and will just add complexity to the recurring code. Its usually Monthly or Annually. I'm open to any input on this. However I do like if a recurring gets missed by cron not running that day, that it generates a new invoice from recurring the next time cron runs

johnnyq:
it might be useful to open a forum discussion on this as I believe this was not discussed, but ideally I would like to keep it Monthly or Annually and eventually just get rid the weekly option

o-psi:
> Im reviewing this and thinking to myself, this isn't typical for business to have recurring charges daily weekly or every 6 monthes etc.

I do not think this is the case. We are in several 3 year contracts. We pay every 6 months to our hosting provider. This may not be the normal, but why not support it. Although I agree daily and weekly is a bit excessive, I can remove them if you wish.

The 1 is just hardcoded into the logic, changing that integer via a variable doesn't really add much complexity... 

I'm open to a forum post, but we need more intervals than 1x, as we have preexisting contracts on that basis, and we are not interested in renegotiating a bunch of our contracts for something as simple as this. 



> eventually just get rid the weekly option

I'm for this. Weekly is never used in business. Monthly, Quarterly, Semi-Annual, Annual are the main ones we have seen.

johnnyq:
Yeah I agree with the Quarterly (3 month) and Semi Annual (6 Month) but lets make it a simple dropdown and remove weekly similar to how @wrongecho programmed the scheduled tickets

See here:
image

johnnyq:
I would say it should be implemented with expenses as well

PR #806, opened 2023-11-14
Add Client-Replied and update select2 statuses in ticket sort dropdown

o-psi:
This pull request adds the ability to filter ticket statuses by any status in the DB via the dropdown. It also updates the select2 multiple statuses to allow for selecting multiple statuses at once. This improves the user experience and makes it easier to find and manage tickets.

### DIFF
--- a/client_assets.php
+++ b/client_assets.php
@@ -130,7 +130,7 @@ $num_rows = mysqli_fetch_row(mysqli_query($mysqli, "SELECT FOUND_ROWS()"));
                                     <a href="?<?php echo $url_query_strings_sort; ?>&type=network" class="btn <?php if ($_GET['type'] == 'network') { echo 'btn-primary'; } else { echo 'btn-default'; } ?>"><i class="fa fa-fw fa-network-wired"></i> Network <span class="right badge badge-light"><?php echo $network_count; ?></span></a>
                                     <?php
                                 }
-                                if ($network_count > 0) { ?>
+                                if ($other_count > 0) { ?>
                                     <a href="?<?php echo $url_query_strings_sort; ?>&type=other" class="btn <?php if ($_GET['type'] == 'other') { echo 'btn-primary'; } else { echo 'btn-default'; } ?>"><i class="fa fa-fw fa-tag"></i> Other <span class="right badge badge-light"><?php echo $other_count; ?></span></a>
                                     <?php
                                 } ?>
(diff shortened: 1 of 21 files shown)


wrongecho:
Test these changes at: https://addclientrepliedtoticketstatussort806.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Hi I like the new status option you added Client Replied but we only want to keep one status at a time, we do not want multiple statuses at any given point.

johnnyq:
Also maybe to identify the last response easier instead of going into the ticket we could use put the person that last responded in the Last Response column under tickets.

o-psi:
The ticket cannot have multiple statuses. That would be dumb. I agree,

This only applies to the filter on the main ticket view screen. That way if you want a screen with any open, in progress and client responded tickets,

johnnyq:
ah I see what you did, i'm doing too many things at once haha this looks good ill reel it in

wrongecho:
I'm a little confused on the benefit of a client replied status. We currently just set the ticket back to open, what's the advantage of separating the two over the simplicity of having a single status for outstanding "agent activity pending" tickets?

o-psi:
@wrongecho No, the ticket gets set to a status of client replied, and I think that is something we shouldn't change.

johnnyq:
Just curious as I have no used many ticketing systems. What do other ticketing system do, do they change the status of the ticket when a client replies?

wrongecho:
I tested this on the demo before writing the message message and when replying on the portal it goes to Open.

johnnyq:
I can see @wrongecho perpective for example if we create a ticket for a client then set the status to pending vendor if the client send an email it will just set it client replied

o-psi:
If the client replied, we need to respond to the client, yes? Our team got used to this and has been using it regularly.

See https://github.com/twetech/itflow/blob/4060bea7a65afb91efb7cc4d7ce2b1a7af50b1c8/cron_ticket_email_parser.php#L314

o-psi:
> Just curious as I have no used many ticketing systems. What do other ticketing system do, do they change the status of the ticket when a client replies?

Zammad, Ninja One, Connectwise Manage, and Kaseya all have a client replied status. Those are the previous ticketing systems we have used.

wrongecho:
> If the client replied, we need to respond to the client, yes? Our team got used to this and has been using it regularly.
> 
> See https://github.com/itflow-org/itflow/commit/ca82a567d703d0045af43b409e1dbb5aeebf439d

I see... I wasn't aware this had been added. It has not been added in the portal.

Regarding needing to respond to the client, it's my view that the Open status shows that ticket action is required but seeing as this has already been shoehorned in already I'll refrain from objecting further.

johnnyq:
Im not ready to jump the PDO bandwagon just yet can we hold off on that

wrongecho:
https://forum.itflow.org/d/465-prepared-statements-by-10

johnnyq:
This looks good other than the PDO, this needs reverted back to mysqli

PR #810, opened 2023-11-17
Enable passing stripe fee to clients

o-psi:
This pull request updates the database version to 0.9.5 and adds a new field called config_stripe_client_pays_fees to the settings table. This field allows the configuration of whether the client should pay the fees for Stripe transactions. This feature is useful for cases where the client wants to cover the transaction fees instead of the business.

### DIFF
--- a/database_updates.php
+++ b/database_updates.php
@@ -1498,12 +1498,12 @@ if (LATEST_DATABASE_VERSION > CURRENT_DATABASE_VERSION) {
 
     }
 
-    //if (CURRENT_DATABASE_VERSION == '0.9.4') {
+    if (CURRENT_DATABASE_VERSION == '0.9.4') {
         // Insert queries here required to update to DB version 0.9.5
-
+        mysqli_query($mysqli, "ALTER TABLE `settings` ADD `config_stripe_client_pays_fees` TINYINT(1) NOT NULL DEFAULT 0 AFTER `config_stripe_account`"); 
         // Then, update the database to the next sequential version
-    //    mysqli_query($mysqli, "UPDATE `settings` SET `config_current_database_version` = '0.9.5'");
-    //}
+        mysqli_query($mysqli, "UPDATE `settings` SET `config_current_database_version` = '0.9.5'");
+    }
 } else {
     // Up-to-date
 }
--- a/database_version.php
+++ b/database_version.php
@@ -5,5 +5,5 @@
  * It is used in conjunction with database_updates.php
  */
 
-DEFINE("LATEST_DATABASE_VERSION", "0.9.4");
+DEFINE("LATEST_DATABASE_VERSION", "0.9.5");
 
--- a/db.sql
+++ b/db.sql
@@ -1267,6 +1267,7 @@ CREATE TABLE `settings` (
   `config_stripe_publishable` varchar(255) DEFAULT NULL,
   `config_stripe_secret` varchar(255) DEFAULT NULL,
   `config_stripe_account` tinyint(1) NOT NULL DEFAULT 0,
+  `config_stripe_client_pays_fees` tinyint(1) NOT NULL DEFAULT 0,
   `config_azure_client_id` varchar(200) DEFAULT NULL,
   `config_azure_client_secret` varchar(200) DEFAULT NULL,
   `config_module_enable_itdoc` tinyint(1) NOT NULL DEFAULT 1,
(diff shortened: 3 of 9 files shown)


wrongecho:
Test these changes at: https://clientpaysfees810.pr-review.itflow.org  
 _(automatic message)_

wrongecho:
Whilst I appreciate the time you've put into this, I do wish you'd raised this as a discussion first as I'm not overly keen on this idea. 

In the UK (consumer market at least), it's unlawful for retailers to specifically charge a fee to pay by card. The business has to eat the cost. I'm not sure if B2B transactions are covered but I'm really not a fan.
https://www.gov.uk/government/news/card-surcharge-ban-means-no-more-nasty-surprises-for-shoppers

johnnyq:
Although it would be neat to automatically create an expense based off the cost of the Stripe Credit Card Transaction, this would save us from manually imputing the expense of every Stripe Transaction

o-psi:
@johnnyq It is **very** common practice for business owners to be willing to pay a fee to use a credit card versus ACH. We have been offsetting the CC fees since day one accepting credit cards. 

I dont know your experience with B2B, but this is extremely commonplace, they do this is because the owner will get cashback on the card that is not subject to Payroll taxes, that they can use for personal expenses. 3/5 of our vendors charge us CC fees, and the president of TWE does not care, as he gets 4% back.


> There are no laws around it in the US at least that I know of

Some states have laws against it, Texas just changed from banning it to allowing it. We started accepting credit cards once we were allowed to offset the fees to the client.
It is up to the business to know the environment in which they are conducting business.


> so why should we enforce this bad behavior

I don't see my role in helping develop ITFlow as in any way to try to control another's business, and I sure hope you agree. We are building features we need. Other businesses may want it as well.
In the end, all other PSA's let you do this. Quickbooks lets you do this. Invoice Ninja lets you do this. 


> if they did I would hop to the next Vendor soo fast

We are not deciding norms here, just features.

And for goodness sake if you looked at the code, it is defaulted to off. The user needs to opt in for it...

wrongecho:
> And for goodness sake if you looked at the code, it is defaulted to off. 

@o-psi, have you reviewed the contributor code of conduct? Whilst I love your passion, your outbursts are growing a little tiring.

o-psi:
> > And for goodness sake if you looked at the code, it is defaulted to off.
> 
> @o-psi, have you reviewed the contributor code of conduct? Whilst I love your passion, your outbursts are growing a little tiring.

On re-reading my comment, I revised it long before you made a statement on it. This is as I recognized it did not carry the meaning intended; regardless, I do not clearly see that I broke the code of conduct. 

I'm glad you mention the code of conduct:
> We pledge to act and interact in ways that contribute to an open, welcoming, diverse, inclusive, and healthy community.
> Being respectful of differing opinions, viewpoints, and experiences
> Giving and gracefully accepting constructive feedback
> Focusing on what is best not just for us as individuals, but for the overall community

My frustration is solely that I feel numerous times there have been either;  A) Strawman fallacies have been constructed in attempts to shut down a PR, or B) There is claimed to be a problem that there is not. I agree the tone in the unrevised comment is a bit harsh, but thus the unrequested revision. 

I am fine with critiques, and I have no issues learning history on the code, but many of the times there has been a "problem", it just needed a closer look at the code. Additionally, there have been many times that a revision is needed, and I am fast to revise with no complaint. 

I am not trying to foster any hard feelings, so let us all put our ego down and work on this project, as we all benefit more from coding than from arguing.

johnnyq:
lets do it, Go!

A Couple of things

- Stripe's flat rate and percentage are set statically in the code these should be grabbed from Stripe before calculating, possibly using their API.
- The Credit Card processing fee should be added as a new line to the customer invoice

o-psi:
> * The Credit Card processing fee should be added as a new line to the customer invoice

Can you cite your source? haha

From my experience, you do not want to account for these fees, unless you are also categorizing them as an expense... Maybe we can incorporate that later, but it is currently in a state that our organization would like to use.

> * Stripe's flat rate and percentage are set statically in the code these should be grabbed from Stripe before calculating, possibly using their API.

I did not see a way to do this, the only way I can see for this is to have a field in the settings where the user can update it. That is how invoice ninja does it aswell.

johnnyq:
The charge needs to show up on the customer's Invoice / Receipt somewhere stating they paid that amount?

johnnyq:
heres an example of what I found when i pulled it up on DDG Image Search
image

johnnyq:
Seems like the picture just answered part of my question. its my misunderstanding on how credit card processing fees work and get calculated here as I have never dealt with anything like this in the past. Im going to assume there is nothing else that needs done here and will just reel this in

o-psi:
@johnnyq I want to add CC fee tracking as an expense. This would also be a checkbox.

PR #814, opened 2023-11-21
Update Array for ticket Query

o-psi:
update the array used for ticket query. As seen in: https://forum.itflow.org/d/561-ticket-filtering-seems-to-be-broken


wrongecho:
Test these changes at: https://fixticketquery814.pr-review.itflow.org  
 _(automatic message)_

PR #819, opened 2023-11-22
Add missing currency variable

o-psi:
This pull request adds a missing currency variable to the client balance report. The currency format is retrieved from the settings using the `getSettingValue` function. This ensures that the client balance is displayed correctly in the desired currency format.


wrongecho:
Test these changes at: https://fixcollectionsformatting819.pr-review.itflow.org  
 _(automatic message)_

PR #827, opened 2023-11-25
Bugfix: New locations are set as the primary

wrongecho:
Bugfix: Every time you add a new location, it is marked as the primary location in the modal by default as the location_primary variable is overwritten with each iteration of a location. This PR defaults to no location being the primary, to avoid unnecessarily overwriting the primary location.

Future work: Re-add fixed logic used to determine if a primary location is missing


wrongecho:
Test these changes at: https://locprimary827.pr-review.itflow.org  
 _(automatic message)_

PR #830, opened 2023-12-04
Fix domain edit for .eu and other domains.

stehled:
Fix for errors from https://forum.itflow.org/d/598-cannot-edit-expiration-date-of-eu-domain

Please comment if there are any issues with the code.


johnnyq:
excellent fix @stehled pulling it in

PR #833, opened 2023-12-13
Statements

o-psi:
Elementary statement generation to be able to send to past due clients.

### DIFF
--- a/client_statement.php
+++ b/client_statement.php
@@ -22,10 +22,10 @@ if (isset($_GET['client_id'])) {
     $result_client_details = mysqli_query($mysqli, $sql_client_details);
     $row_client_details = mysqli_fetch_assoc($result_client_details);
 
-    $client_name = $row_client_details['client_name'];
-    $client_type = $row_client_details['client_type'];
-    $client_website = $row_client_details['client_website'];
-    $client_net_terms = $row_client_details['client_net_terms'];
+    $client_name = nullable_html_entities($row_client_details['client_name']);
+    $client_type = nullable_html_entities($row_client_details['client_type']);
+    $client_website = nullable_html_entities($row_client_details['client_website']);
+    $client_net_terms = intval($row_client_details['client_net_terms']);
 
     $sql_client_unpaid_invoices = "
     SELECT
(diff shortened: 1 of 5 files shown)


wrongecho:
Test these changes at: https://statements833.pr-review.itflow.org  
 _(automatic message)_

o-psi:
@johnnyq

johnnyq:
Looks good so far! however a few things...

- the vars will need to be escaped to prevent XSS
- I'm debating if I like the idea that statements should have its own side menu or maybe be inside payments sub nav, what are your thoughts on this @o-psi

o-psi:
@johnnyq I like the sub menu idea. We will access it via the links on the collections report. 

Its hard for me to escape while I'm working on the logic for some reason, usually do it last. Must have forgot; as you can tell we are in a bit of a rush haha. Our accounts payable is sitting at 3x MRR 🙄 after the switch to ITFlow, so I guess our adoption is lower than expected.

I also want to in the future add the company logo, and make it accessible via the guest view.

I will work on that shortly!

johnnyq:
ah gotchya that makes sense, keeps the code clean for focus, sounds good =]

functions.php
```diff
@@ -870,5 +870,24 @@ function calculateAccountBalance($mysqli, $account_id) {
...
     }
 
+    return $balance;
+}
+
+function calculateInvoiceBalance($mysqli, $invoice_id) {
+    $sql_invoice = mysqli_query($mysqli, "SELECT * FROM invoices WHERE invoice_id = $invoice_id");
+    $row = mysqli_fetch_array($sql_invoice);
+    $invoice_amount = floatval($row['invoice_amount']);
+    $invoice_id = $row['invoice_id'];
```
johnnyq, line 880: This one needs sanitized with intval

o-psi:
@johnnyq Think were good to go here. I looked over all the variables once more.

client_statement.php
```diff
@@ -0,0 +1,188 @@
...
+                    <tbody>
+                        <?php 
+                            while ($row = mysqli_fetch_assoc($result_client_unpaid_invoices)) {
+                                $invoice_number = intval($row['invoice_number']);
+                                $invoice_id = intval($row['invoice_id']);
+                                $invoice_prefix = nullable_html_entities($row['invoice_prefix']);
+                                $invoice_date = nullable_html_entities($row['invoice_date']);
+                                $invoice_amount = floatval($row['invoice_amount']);
+                                $invoice_amount_formatted = numfmt_format_currency($currency_format, $invoice_amount, $currency_code);
+                                $invoice_url = intval("invoice.php?invoice_id=$invoice_id");
```
johnnyq, line 91: for this one remove intval() it will turn the whole url into a number invoice_id is already esscaped so its not needed for this line

isaacb82: @johnnyq this is now fixed.

johnnyq:
alright looks good, i'm gonna go ahead and merge it.
I like that debug section. I almost thing that shoould be settings like config_debug_enable and if it == 1 then include it.

o-psi:
@johnnyq It is the plan, to include it everywhere if in setting debug is turned on, via inc_all.php

PR #834, opened 2023-12-17
Update ticket billing functionality

o-psi:
This pull request includes updates to mark tickets as billable and keep track of which tickets have been billed. It also adds the ability to link tickets to invoices and vice versa.

### DIFF
--- a/ticket_invoice_add_modal.php
+++ b/ticket_invoice_add_modal.php
@@ -1,3 +1,8 @@
+<?php
+// Check if ticket_id and invoice_id are set in the URL
+$addToExistingInvoice = isset($_GET['ticket_id']) && isset($_GET['invoice_id']);
+?>
+
 <div class="modal" id="addInvoiceFromTicketModal" tabindex="-1">
     <div class="modal-dialog">
         <div class="modal-content bg-dark">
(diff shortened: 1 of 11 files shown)


wrongecho:
Test these changes at: https://billabletickets834.pr-review.itflow.org  
 _(automatic message)_

o-psi:
@johnnyq thoughts?

johnnyq:
@o-psi something broke tinymce here in ticket.php take a peek see if you can find what may have broke it.
Also we do not like to use PHP short hand as it doesnt explain the code very well.
but other than that this is very exciting,
ill review more once those are fixed.

o-psi:
Strange thing, TinyMCE works fine in my dev env, wonder if it is due to the new open_modal javascript? Is that possible with TinyMCE?

I will work on the shorthand, I assume you mean the ?? ::

ticket_invoice_add_modal.php
```diff
@@ -13,26 +18,26 @@
                     
                     <ul class="nav nav-pills nav-justified mb-3">
                         <li class="nav-item">
-                            <a class="nav-link active" data-toggle="pill" href="#pills-create-invoice"><i class="fa fa-fw fa-check mr-2"></i>Create New Invoice</a>
+                            <a class="nav-link <?php echo !$addToExistingInvoice ? 'active' : ''; ?>" data-toggle="pill" href="#pills-create-invoice"><i class="fa fa-fw fa-check mr-2"></i>Create New Invoice</a>
```
o-psi, line 21: Is this the PHP Shorthand in question?

johnnyq: yes thats it

johnnyq:
what is the show modals js code used for?

ticket_invoice_add_modal.php
o-psi: @johnnyq This has been fixed

johnnyq:
ah gotchya okay that makes sense.
Still reviewing the code but its looking promising much needed feature

johnnyq:
looks good im gonna reel in and we'll fix thing as they arise if any issues are there.

PR #839, opened 2023-12-22
Fix email data assignment and select option bug in invoice.php and ticket_invoice_add_modal.php

o-psi:
This pull request fixes the bug in invoice.php and ticket_invoice_add_modal.php where the email data assignment was incorrect and the select option was not being displayed correctly. The bug caused the email receipt to not be sent and the invoice select option to not be selected properly. This PR corrects the assignment of the email data and ensures that the select option is displayed correctly.

### DIFF
--- a/post/invoice.php
+++ b/post/invoice.php
@@ -686,14 +686,14 @@ if (isset($_POST['add_payment'])) {
                     'body' => $body
                 ];
 
-                $email_data = $email;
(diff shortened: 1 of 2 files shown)


wrongecho:
Test these changes at: https://CreatePaymentBugfix839.pr-review.itflow.org  
 _(automatic message)_

PR #840, opened 2023-12-22
Fix ticket_assigned_to filter in SQL query

o-psi:
This pull request fixes the ticket_assigned_to filter in the SQL query. Previously, the filter was using the LIKE operator with a wildcard at the beginning and end of the filter value. This caused incorrect results when searching for specific users, as you would want. The fix removes the wildcard at the beginning and end of the filter value, ensuring accurate filtering based on the assigned_to field.

### DIFF
--- a/tickets.php
+++ b/tickets.php
@@ -1,5 +1,6 @@
 <?php
 
+
 // Default Column Sortby Filter
 $sort = "ticket_number";
 $order = "DESC";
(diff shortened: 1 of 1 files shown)


wrongecho:
Test these changes at: https://FixforuserIDsintickets840.pr-review.itflow.org  
 _(automatic message)_

o-psi:
I am user ID 1, any technician that has a 1 in their ID shows up in my "My Tickets", so 10-19, 21, 31....

This fixes that issue.

PR #855, opened 2024-01-03
Update settings_backup.php

aftechro:
Create and restore backups from backup list. To do: restore from file


wrongecho:
Test these changes at: https://patch14855.pr-review.itflow.org  
 _(automatic message)_

PR #870, opened 2024-01-22
Add Let's Encrypt Legal Notice

o-psi:


wrongecho:
Test these changes at: https://letsencryptlegalnotice870.pr-review.itflow.org  
 _(automatic message)_

wrongecho:
My intention is not to criticize you or the valuable contributions in this PR. However, I believe it might be more prudent to incorporate the message directly into the install script, as other locations may mention on this script without including the notification.

o-psi:
Love that fantastic idea, definitely an oversight on my end, with less maintenance overall.

PR #875, opened 2024-02-08
0.1.8.2 (Scheduled tickets, and Email Improvements)

o-psi:
This is the most up to date version of the scheduled tickets change. Added iCal Functionality via The Zap Cal PHP Library


wrongecho:
Test these changes at: https://0182875.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Wow that ical support looks super exciting, I have been wanting something like that, Im gonna review then we'll pull it in

johnnyq:
For the iCal functionality I don't think we need to store it in the DB as a separate field It should just send as an attachment, correct?

o-psi:
I was debating this, but I want the iCal to also show in the mail queue as well eventually, and I don't see that happening easily on the fly any other way.

o-psi:
Also, If the ticket gets rescheduled, a new email needs to be sent to replace the old ICS, this would best be done by comparing when that field changes.

johnnyq:
gotchya that makes sense to me

o-psi:
I am going to work on this a bit more. The emails need some help, and there should be a note in the ticket when scheduled.

johnnyq:
okay sounds good its looking great so far

o-psi:
@johnnyq I think this is ready to go

johnnyq:
looks good, im, reeling in

o-psi:
@johnnyq I would love to write some docs on how this functionality works, yet it seems I do not have perms to create new pages on docs.itflow.com

PR #877, opened 2024-02-12
Enhance ticket reply button

wrongecho:
- Better logic on when the button should default to private/public
- Fix the new plane icon not showing when the button was checked/unchecked


wrongecho:
Test these changes at: https://ticketrespondbehaviour877.pr-review.itflow.org  
 _(automatic message)_

PR #882, opened 2024-02-15
0.1.8.4 (Recurring client sidebar)

o-psi:


wrongecho:
Test these changes at: https://0184882.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Nice work on all this, 
And just brain storming here but 
I was thinking that we should probably have a separate table called event_attendees

attendee_id
attendee_name
attendee_email
attendee_notes
attendee_declined_at
attendee_accepted_at
attendee_created_at
attendee_updated_at
attendee_contact_id --- If one of the contacts is selected then reference the contact_id for linkage later on maybe in reporting or history tracking etc otherwise this can happily default to 0
attendee_event_id

 This is so each event can have multiple attendees. I didn't add a archived_at as it doesn't matter if its deleted

Let me know your thoughts and or if there is anything I missed or you feel should be different

Wanted to also see if you wanted to take a stab at it since your focusing on calendar events at the moment?

o-psi:
i like the idea, But we are not using events any, so I have yet to figure them out haha.

PR #885, opened 2024-02-22
Avoiding issue when unable to add domain

NickyM:
We were experiencing HTTP 500 when trying to add the following domains:

- solemaids.nl
- solemaids.com.au

The getDomainExpirationDate function returned NULL. The mySQL column isnt nullable.
Returning mySQL min date instead.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -373,10 +373,11 @@ function encryptLoginEntry($login_password_cleartext)
 // Get domain expiration date
 function getDomainExpirationDate($name)
 {
-
+    $nullValue = "1000-01-01";
+    
     // Only run if we think the domain is valid
     if (!filter_var($name, FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME)) {
-        return "NULL";
+        return $nullValue;
     }
 
     $ch = curl_init();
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -373,10 +373,11 @@ function encryptLoginEntry($login_password_cleartext)
 // Get domain expiration date
 function getDomainExpirationDate($name)
 {
-
+    $nullValue = "1000-01-01";
```
o-psi, line 376: This should be 1 January 1970 00:00:00, the start of the current epoch

o-psi (changes requested):
Other than the date, this looks great. Should help with the forum posts as well.

functions.php
NickyM: I just took mySQL min date, but thats an easy change.

Preferable it should be nullable, but I have no clue of any potential issues in the codebase if the value is null. Just changed that few lines as it was stopping me from moving our docs :D

o-psi: I do not know the domain portion of this tool enough to say exactly. I think utc 0 is fine for now.

johnnyq:
Hey @NickyM Thank you for the pull request however this should remain NULL as this will throw off our expire checks that we have throughout the code and will produce some unintended results. Basically the expire date should be NULL if no date is given.

NickyM:
@johnnyq - if it's NULL the server throws a HTTP 500 because the mySQL table isn't nullable.

wrongecho:
Looking into this further, just running a command-line `whois` against `.be`, `.nl` and `.com.au` domains, I can't even see a whois that is returned? Doesn't seem these not publicly shown for these TLDs, so will not look into any further at this time.

--

I've raised a separate [PR](https://github.com/itflow-org/itflow/pull/888) to properly account for null values. Very similar to your PR @NickyM, but has some improved logic and should cover the cron bits off too.

If you get a moment, please play around on the PR Review page and see if you can break it?

PR #888, opened 2024-02-22
Domain expiration dates

wrongecho:
- Update logic in post when adding/editing a domain to better account for null values
- Update logic in cron domain refresher to account for null values
- Prevent cron domain refresher getting stuck on a single domain
- Exclude domains with no expiration date from the cron nightly renewal alerts

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -138,7 +138,7 @@ if($config_enable_alert_domain_expire == 1){
             $mysqli,
             "SELECT * FROM domains
             LEFT JOIN clients ON domain_client_id = client_id 
-            WHERE domain_expire = CURDATE() + INTERVAL $day DAY"
+            WHERE domain_expire IS NOT NULL AND domain_expire = CURDATE() + INTERVAL $day DAY"
         );
 
         while ($row = mysqli_fetch_array($sql)) {
--- a/cron_domain_refresher.php
+++ b/cron_domain_refresher.php
@@ -41,12 +41,20 @@ if ( $argv[1] !== $config_cron_key ) {
 // END ERROR
 // REFRESH DOMAIN WHOIS DATA (1 a day)
 //  Get the oldest updated domain (MariaDB shows NULLs first when ordering by default)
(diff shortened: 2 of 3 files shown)


wrongecho:
Test these changes at: https://dnsexpirynullfix888.pr-review.itflow.org  
 _(automatic message)_

NickyM:
Pardon me for not having too much insight into either PHP or ITFlow, but shouldnt there be an ALTER TABLE statement somewhere? And changes to database initializaiton?

wrongecho:
@NickyM I thought the same, but it turns out we actually already allow NULL for the domain expiry field!

 https://github.com/itflow-org/itflow/blob/31184f0db19eeaade58086abab478e2b3b959bf3/db.sql#L504

I _think_ the issue was the code previously was trying to add it as a variable which was then enclosed in quotes in the SQL statement, so it was treated as a string rather than an actual null.

NickyM:
Just realised that the error message that I got yesterday, surelay states 'NULL' and not NULL for the error. :)

> [Thu Feb 22 08:47:47.736482 2024] [php:error] [pid 2047] [client 172.29.0.1:48080] PHP Fatal error:  Uncaught mysqli_sql_exception: Incorrect date value: 'NULL' for column `itflow`.`domains`.`domain_expire` at row 1 in /var/www/html/post/domain.php:40\nStack trace:\n#0 /var/www/html/post/domain.php(40): mysqli_query()\n#1 /var/www/html/post.php(38): require_once('...')\n#2 {main}\n  thrown in /var/www/html/post/domain.php on line 40, referer: [removed]

The domain adding/editing works fine for me then.

johnnyq:
@wrongecho @NickyM Things are working fine for me too. I feel confident to reel in.

wrongecho:
@NickyM 
> Just realised that the error message that I got yesterday, surelay states 'NULL' and not NULL for the error. :)

It's truly amazing how tiny, insignificant errors can make or break big features haha! Thanks again for your help on this - has been a thorn in my side since it was first added! :)

PR #918, opened 2024-03-20
0.2.0.0 Make POST actions use functions that API V2 can use

o-psi:
{WIP}

This should simplify the codebase with a branch in mind, and makes the API more maintainable.


wrongecho:
Test these changes at: https://0200functionalizing918.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Hey @o-psi there quite a bit to go through here. It will take me a few weeks to test and go through properly. We already have an API in place, although it could use additional end points as you stated. Also I don't want to change the POST code into functions. I want to keep the code style the way it is. So if you remove all those POST functions ill review the code again.

johnnyq:
Also Can we put each small feature into its own pull request in the future as well. There no way my brain can not process this in a timely manor as i'm also working on core code duties. I'm not able to work efficiently this way.

o-psi:
@johnnyq I think this may be our fork in the road.

For the version of this project I am maintaining, I will definitely use functions instead of hard coding it all.

Issue #925, opened 2024-03-24
Feature Request: Custom Fields - Tickets - Ticket Status


wrongecho:
Some thoughts on how to allow users to add and use custom ticket statuses...

--

New table `ticket_statuses`
- ticket_status_id - int
- ticket_status_name - varchar for the name
- ticket_status_color - color for the ticket status (bootstrap: primary, success, danger, dark)
- ticket_status_active - 0/1, default 1 / active & visible
_Possibly add extra fields in future to allow for SLA pausing, etc_

Existing table `tickets`
- ticket_id
- [...]
- ticket_status - Use the ticket_status_id here

Functions
- Will also need to adjust functions.php - `getTicketStatusColor()` to allow looking up the correct colour
- Possibly add a custom function to automatically lookup the ticket status rather than having to add a join everywhere? `getTicketStatusName()` ?

marnel007:
This is the one thing holding us back from moving to ITFLOW, We have build our own statues code,  It would be nice..

![image](https://github.com/itflow-org/itflow/assets/61925531/cff2e0cd-a061-47b7-a3ee-e65cecaff479)

wrongecho:
Hey @marnel007 
Custom ticket statuses are literally just around the corner!

marnel007:
Can you tell me how "around the corner"  week / month / days ....

johnnyq:
days

marnel007:
Bug in picking colors 

![image](https://github.com/itflow-org/itflow/assets/61925531/aff3b6d8-940f-4e5f-a7eb-0a00270d65ba)

johnnyq:
Hey @marnel007 almost there this will be complete by end of today

johnnyq:
Closing as this is completed

marnel007:
Thank you !!

PR #926, opened 2024-03-24
Ticket Statuses from DB

wrongecho:
First swing at this to share my progress, isn't ready to merge yet but would appreciate thoughts and any help you can offer - it's a lot more work than I anticipated due to filters. 
Added default status values and some new config fields too as we'll need a way to reference what New/Open/Closed means (as people might want to change this in future)

image

image

image

Functions
- getTicketStatusColor() -- Gets colours for badges (bootstrap)
- getTicketStatusName() -- Resolves a ticket_status_id to a readable name (for GUI/Emails)

https://github.com/itflow-org/itflow/issues/925


wrongecho:
Test these changes at: https://ticketstatuses926.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
This looks fantastic and I do like the idea of using functions to get ticket status name and color. I think this is deff a good start I do not see anything wrong with pushing this into play

Issue #937, opened 2024-04-07
Feature Request: Client contacts ticket notification outside of original ticket creation flow


wrongecho:
_Related to support req TCK-20._

If a ticket is created with nobody as the contact, or the contact/company is changed, then the new contact is never notified the ticket exists.
We need a way to email the latest contact of a ticket the original "New ticket raised" email - either intelligent logic when the contact is changed or a button to manually send the email.

compumatter:
Thank you for breathing life into it.  Appreciated!

PR #943, opened 2024-04-13
Ticket - Notifications

wrongecho:
Allow notifying contacts & watchers of a ticket when they are added/updated (#937)

image

image

_(The checkbox will be pre-ticked on the Contact change modal if "Send clients general notification emails" is turned on. This isn't the case on the edit ticket detail as that would likely cause lots of emails.)_


compumatter:
Thank you for adding this functionality!

PR #947, opened 2024-04-18
Update side_nav.php

git-kup:
Fix all 3 "Recurring" buttons where not inline with the rest of the buttons in the main side nav. it was slightly indent to the right.


johnnyq:
Thanks

PR #948, opened 2024-04-27
Enable URL Recovery from logout

ssteeltm:
to acomplish the discussed in https://forum.itflow.org/d/688-session-expire-should-back-with-current-url

When yout get logged out from session expire, it will save last url tried, to when logging again redirect to it.

### DIFF
--- a/check_login.php
+++ b/check_login.php
@@ -18,7 +18,10 @@ if (!isset($config_enable_setup) || $config_enable_setup == 1) {
 
 // Check user is logged in with a valid session
 if (!isset($_SESSION['logged']) || !$_SESSION['logged']) {
-    header("Location: login.php");
+    if($_SERVER["REQUEST_URI"] == "/")
+        header("Location: login.php");
+    else
+        header("Location: login.php?url=".urlencode($_SERVER["REQUEST_SCHEME"] . "://" . $_SERVER["HTTP_HOST"] . $_SERVER["REQUEST_URI"]) );
     exit;
 }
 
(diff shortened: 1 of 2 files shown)


wrongecho:
Hey there,

Thanks for the PR!
Been super busy so I haven't had a chance to try this out yet, but I like the idea!

My only issue is that if you put an external link in the URL parameter, it looks like you could potentially divert people to third party websites. This is known as an Open Redirect vulnerability and is generally used in phishing campaigns. 

https://portswigger.net/kb/issues/00500100_open-redirection-reflected

https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html


I'm sure there's a way to mitigate but would love to hear your thoughts?

ssteeltm:
I see now the potential issue.
What if, we just save the uri, since domain we can get from config, and also, we could hash somehow the query parameter

wrongecho:
This looks good to me!

The only thing I could think may be an issue is setting the location based off the `$_SERVER["REQUEST_SCHEME"]`. I know some will be running ITFlow HTTP (as far as the server is concerned) but then using a reverse proxy/tunnel for the HTTPS. What do you think about defaulting that to HTTPS or `//` (which should default to whatever the users browser is connecting with)?

 _(Nitpick: would be fab if you could add a space after the IF statement and include braces for both - it's our standard, even if there's only one line within the statement)_.

Either way @johnnyq I'm happy to get this merged for now and fix forward.

wrongecho (approved):

johnnyq:
Looks great to me @wrongecho and thank you for the patch @ssteeltm  merging in

PR #954, opened 2024-05-06
Phone mask option

ssteeltm:
Enable or Disable Phone Number Formating ( mask ), proposed on https://forum.itflow.org/d/39-regional-formats-settings/23


johnnyq:
Looks great, you even handled the Database updates quite well and right in line with ITFlow's coding style great job

PR #957, opened 2024-05-13
search in all folders when in root (docs/files)

ssteeltm:
Enable search recursive/all folders when in / (root).

In the future will be nice to have this as an advanced search instead.

reference:
https://forum.itflow.org/d/850-search-documents-recursive


johnnyq:
Looks good, reeling in Thanks @ssteeltm

PR #958, opened 2024-05-16
client_logins uri limit with ellipsis

ssteeltm:
limits long uri that may break lines in table

example:
image

it put ellipsis on the displayed  long uri


johnnyq:
nice work and deff needed for those long URIs however we have a PHP function for this already in the functions.php
```
function truncate($text, $chars)
{
    if (strlen($text) <= $chars) {
        return $text;
    }
    $text = $text . " ";
    $text = substr($text, 0, $chars);
    $lastSpacePos = strrpos($text, ' ');
    if ($lastSpacePos !== false) {
        $text = substr($text, 0, $lastSpacePos);
    }
    return $text . "...";
}
```

Please update your pull request to use the function, and thank you for your commitment to ITFlow

johnnyq:
awesome looks good reeling in

PR #960, opened 2024-05-17
Set 2FA Remember-me cookie expiry

wrongecho:
Currently, the token is only valid for 2 days (86400 seconds = 24 hrs, multiplied by 2). 

This PR adjusts the cookie expiry date to the number of days configured that tokens are cleared after. This should help ensure users are not prompted for 2FA every few days, even if they've set a longer interval.


johnnyq:
Good catch @wrongecho Thanks for the quick fix =]

PR #961, opened 2024-05-26
2FA QR Codes

wrongecho:
Move away from the Google Charts API for generating QR codes as this is currently broken and not a security best practice.

Note: Generating images locally requires the 'gd' extension for PHP. Docs and install script has been updated.


wrongecho:
Test these changes at: https://qrcodefix961.pr-review.itflow.org  
 _(automatic message)_

PR #964, opened 2024-05-26
User roles - Initial

wrongecho:
Move the default user roles into the database to make way for custom roles & permissions.

Please could you take a look, test, and let me know your thoughts? 
Following from this, I'd like to create a role_permissions table where we then re-create the default permissions that we have in the code in the database. Once all that is complete, we can then open up to editing the built-in roles / allowing custom roles and permissions.

### DIFF
--- a/admin_user_add_modal.php
+++ b/admin_user_add_modal.php
@@ -55,9 +55,15 @@
                             </div>
                             <select class="form-control select2" name="role" required>
                                 <option value="">- Role -</option>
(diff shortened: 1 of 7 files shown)


wrongecho:
Test these changes at: https://userrolesinitial964.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
This looks great!! Code is clean, can't seem to find any where this might break pulling in

PR #970, opened 2024-06-04
use building icon

git-kup:
use building icon instead of users icon as it's more relevant
if you don't agree, disregard the PR... :)


wrongecho:
Test these changes at: https://patch2970.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Hey @git-kup Thank you for the pull request but I will have to kindly disagree as we user the users icon everywhere for clients. The Building is reserved for the vendors

PR #972, opened 2024-06-09
2FA tweaks

wrongecho:
- Set the 2FA number input field to only accept 6 characters max
- Revoke existing remember-me tokens when 2FA is re-enabled


wrongecho:
Test these changes at: https://2fa6char972.pr-review.itflow.org  
 _(automatic message)_

PR #973, opened 2024-06-09
Email to ticket parsing - Save the original email

wrongecho:
When creating a new ticket, save the original email in a .eml format and attach it to the ticket.

image

This will be useful for situations when you want to reply to the original thread but don't have access to the inbox. We could add this to all received emails if there is interest?


wrongecho:
Test these changes at: https://ticketparsesaveoriginalmail973.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Very nice addon, good work!

PR #974, opened 2024-06-14
Update client_file_link_asset_modal.php

ssteeltm:
small adjustment to allow link an asset in any files. Before this it was limiting to link only one file, dont displaying it in assets listing for other files


johnnyq:
Thank you @ssteeltm

PR #976, opened 2024-06-30
Add audit log cleanup/retention period to cron

wrongecho:
Audit logs will be automatically cleaned up after 90 days (new installs) or 7 years (existing installs, in-case there is a compliance requirement for some of our existing users). 

This is configurable in Settings > Security.

image


wrongecho:
Test these changes at: https://logcleanup976.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
nicee!!

PR #978, opened 2024-06-30
Email Template Editor

jack-10000:
I have added an email template editor with TinyMCE to Admin > Settings > Mail.
This allows an admin to edit the email templates for all customer-facing email bodies and subjects in ITFlow.

Shortcode support allows you to add things like [company_name] or [ticket_prefix] to insert those values into your template body or subject. You can also insert something like [[ticket_prefix][ticket_number]] or [[company_name]] too.
Shortcodes are converted to $variables if they match a predefined list in the _prepareEmailTemplateTags_ function in functions.php.

There is an 'email-templates-sample.sql' file containing sample templates that are ready to use and will update the DB structure necessary.


johnnyq:
Thanks for doing this however we will be taking a different approach to this in the future

Heart1010:
@johnnyq which approach is planned 😉?

PR #983, opened 2024-07-20
Allow publishing certain documents to client portal

jackveney:
Allows for publishing certain documents to the client portal. Defaults documents to not publish to client portal.

### DIFF
--- a/client_document_add_from_template_modal.php
+++ b/client_document_add_from_template_modal.php
@@ -26,7 +26,7 @@
                   $document_template_name = nullable_htmlentities($row['document_name']);
 
                 ?>
-                <option value="<?php echo $document_template_id ?>"><?php echo $document_template_name; ?></option>
+                  <option value="<?php echo $document_template_id ?>"><?php echo $document_template_name; ?></option>
                 <?php
                 }
                 ?>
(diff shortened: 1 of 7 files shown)


wrongecho:
Hey @jackveney

Thanks for the PR! Sorry for the delay in coming back to you.

Not sure if you're aware, but currently all documents are visible to technical client contacts within the portal as we have the idea that the documents belong to the clients. The only downside to this is that there's no way to publish selected documents to all contacts, e.g. KB articles (personally I'd like a separate module just for this but who knows). 
Is this the challenge this PR is trying to solve? 

If we were to move forward with this, it's worth mentioning that to make database changes you need to adjust both db.sql (new installs) and database_updates.php (for existing installs). Also, no changes have been requested to portal/documents.php (or document.php) which would allow the client contacts to actually see the documents.

Thanks in advance!

wrongecho:
Soft closing for now, feel free to reopen.

wrongecho:
@jackveney - FYI that this will become a feature via https://github.com/itflow-org/itflow/pull/1035
Default will be that docs _are_ published to the portal. 

Please feel free to contribute further via PRs or on the Forum :)

PR #991, opened 2024-08-10
Add move/transfer assets between clients

wrongecho:
Add ability to move / transfer assets between clients. Moving an asset involves the asset being archived and then re-created under the new client with _most_ of the same information.

Information about the old & new asset is stored in the audit logs and within the asset notes. Information from the audit logs is accessible via a new History tab on the asset, subject to audit log retention.

image

image

image


wrongecho:
Test these changes at: https://assetclienttfr991.pr-review.itflow.org  
 _(automatic message)_

PR #993, opened 2024-08-15
Only show most current version of a document in the client portal

wrongecho:
Currently all revisions are shown in the client portal, which is confusing.


wrongecho:
Test these changes at: https://portaldocsshowcurrent993.pr-review.itflow.org  
 _(automatic message)_

PR #1001, opened 2024-08-19
Bugfix: credentials

wrongecho:
Fix an edge-case bug causing the user_encryption_session_key session cookie to not be set due to error output (when display PHP errors in browser is enabled). This means login credentials are still encrypted but cannot be decrypted properly by other users. 

As a failsafe, prevent users creating new credentials if they do not have the correct cookie set.

_This is the first time in 2+ years I've run into this, and only because I've setup a new dev env. I don't think it's a common issue by any means but doesn't hurt to fix._


wrongecho:
Test these changes at: https://credfix1001.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
nice catch, yeah deff a rare RACE condition

PR #1032, opened 2024-09-06
Automatic redirect to HTTPS

k1ngErr0r:
Had an issue were - if user manually enters http in URL they get error. 
Code change automatically redirects any http request to https - this should avoid the error.

### DIFF
--- a/login.php
+++ b/login.php
@@ -1,5 +1,16 @@
 <?php
 
+// Force HTTPS redirection if accessed over HTTP or forwarded from HTTP
+if (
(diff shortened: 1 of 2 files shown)


wrongecho:
Hey there @k1ngErr0r
Thanks for the PR.

We decided a while ago that this should be handled at the server level (Apache/Nginx) rather than within the app. 

This is because you may have a perfectly valid reason to serve HTTP (e.g. Running ITFlow behind a forward proxy / load balancer that handles the SSL offloading).

k1ngErr0r:
Makes sense - change works for my case but would work behind forward proxy too.  

Thanks :)

PR #1056, opened 2024-09-20
Client import - resiliency

wrongecho:
- Add some resiliency to the client import function to better account for blank fields (may also fix some import errors)
- Fix the default settings page not loading due to the removal of account types


wrongecho:
Test these changes at: https://clientimport1056.pr-review.itflow.org  
 _(automatic message)_

PR #1058, opened 2024-09-20
Update settings_online_payment.php

ThaMunsta:
Fixed SQL and removed account type condition as I understand that isn't used anymore.

calculateAccountBalance() in functions.php references account_types still but that function seems unused at the moment.


wrongecho:
Test these changes at: https://settingsonlinepaymentfix1058.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Good catch ;-)

PR #1060, opened 2024-09-20
Update dashboard.php to add some more data

jaytagdamian:
The dashboard appears completely empty when ticketing and accounting modules are disabled.

Modified the page to add some basic dashboard elements (recently updated clients, technical dashboard sans ticketing).

image

image


jaytagdamian:
This is almost a straight copy of the code from client_overview.php, no <th> required in the table.

PR #1061, opened 2024-09-20
Recurring invoices - hide delete when active

wrongecho:
Adjust recurring invoices so they have to be made inactive to delete them


wrongecho:
Test these changes at: https://recdelete1061.pr-review.itflow.org  
 _(automatic message)_

PR #1067, opened 2024-09-21
Detect and convert non-UTF8 encoding as part of input sanitization

wrongecho:


wrongecho:
Test these changes at: https://encoding1067.pr-review.itflow.org  
 _(automatic message)_

PR #1071, opened 2024-09-26
Update expenses.php

ThaMunsta:
Downloaded filename extension was hardcoded for PDF and downloads any jpg/png as a PDF file. This fix takes into account the database filename and extension and updated icon to general file.


wrongecho:
Test these changes at: https://downloadfilenamefix1071.pr-review.itflow.org  
 _(automatic message)_

PR #1074, opened 2024-09-29
Refactor POST handling

wrongecho:
- Split into admin and user handlers, each admin page gets its own file now
- Enforce role access once for admin POST requests
- Automatically load POST logic for admin-based requests based on the referring page, otherwise automatically load all user request logic
- Add support for using custom POST handlers


wrongecho:
Test these changes at: https://moreperms1074.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
Nice Work!! @wrongecho

PR #1076, opened 2024-09-30
cron_ticket_email_parser.php - send new ticket opened email to unknown clients

cybershieldms:
- Will now send unknown client (client_id = 0) a ticket open notification email 
- Added regex to not parse bounce e-mails (e.g. From: postmaster, mailer-daemon)
- Fixed ticket description missing from new ticket notifications sent to others.
- Added E-Mail From line display name to beginning of ticket description/e-mail quote with e-mail address and date.

### DIFF
--- a/cron_ticket_email_parser.php
+++ b/cron_ticket_email_parser.php
@@ -87,7 +87,7 @@ function addTicket($contact_id, $contact_name, $contact_email, $client_id, $date
     $message = nl2br($message); // Convert newlines to <br>
     
     // Wrap the message in a div with controlled line height
-    $message = "<i>Email from: $contact_email at $date:-</i> <br><br><div style='line-height:1.5;'>$message</div>";
+    $message = "<i>Email from: <b>$contact_name</b> &lt;$contact_email&gt; at $date:-</i> <br><br><div style='line-height:1.5;'>$message</div>";
 
     $ticket_prefix_esc = mysqli_real_escape_string($mysqli, $config_ticket_prefix);
     $subject_esc = mysqli_real_escape_string($mysqli, $subject);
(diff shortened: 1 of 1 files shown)


wrongecho:
Test these changes at: https://ticketfeatures1076.pr-review.itflow.org  
 _(automatic message)_

cron_ticket_email_parser.php
```diff
@@ -475,9 +477,11 @@ function getInboxFolder($client, $inboxNames) {
                     }
                 } elseif ($config_ticket_email_parse_unknown_senders)  {
                     // Parse even if the sender is unknown
-
-                    if (addTicket(0, 'Guest', $from_email, 0, $date, $subject, $message_body, $message->getAttachments(), $original_message_file)) {
-                        $email_processed = true;
+		    $bad_from_pattern = "/daemon|postmaster|reply|root|admin/i";
```
wrongecho, line 480: I like this concept, but would probably omit 'reply' and 'admin'. 

We probably want automated emails from things like noreply@microsoft.com to be parsed

Some clients use admin@ for shared mailboxes.

wrongecho:
Hey @cybershieldms 

Thanks for the PR! :)

I've tested and this seems to work nicely. I like the way you've put the persons name in bold and email in brackets too!

image

Great concept with excluding bounces but I'd suggest we exclude admin/reply as these can often be used legitimately/for messages you want.

cron_ticket_email_parser.php
cybershieldms: I agree.  I could see actually making the block list regex a configurable field later on in Settings.  Might take a stab at that later on.  I'll fix it now to just exclude daemon and postmaster.

cybershieldms:
> Great concept with excluding bounces but I'd suggest we exclude admin/reply as these can often be used legitimately/for messages you want.

@wrongecho - Thanks!  First time with Github.   Done... Changed the regex to just include postmaster and daemon.

In the future, I could see a configurable variable in Settings to make a customizable block list.

wrongecho:
> @wrongecho - Thanks!  First time with Github.   Done... Changed the regex to just include postmaster and daemon.
> 
> In the future, I could see a configurable variable in Settings to make a customizable block list.

You've done great for your first time!! :)

Agreed, it would be nice to make the ignore list configurable.

wrongecho (approved):

johnnyq:
awesome, we'll pull this one in

wrongecho:
Nice work Craig!! :)

PR #1079, opened 2024-10-02
Add custom event handler 

wrongecho:
Some events in post.php and across the app (like raising a ticket) will now call an event handler. This will allow you to write your own PHP code to be run when certain things in ITFlow happen. 

The event handler will be passed the action that triggered them as $trigger (e.g. ticket_create) and the associated entity ID number as $entity (e.g. 145).

The event handler function checks for the presence of a PHP file in the xcustom directory called xcustom_action_handler.php and includes it in the page if it exists. The custom PHP code in xcustom_action_handler.php will be able to access the $trigger and $entity alongside the ITFlow database to perform the relevant lookups (e.g. checking the ticket contact & grabbing their mobile) and sending them a text via a third party.

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -152,7 +152,7 @@ if ($config_whitelabel_enabled && !validateWhitelabelKey($config_whitelabel_key)
 
 // DOMAINS EXPIRING
 
-if($config_enable_alert_domain_expire == 1){
+if ($config_enable_alert_domain_expire == 1) {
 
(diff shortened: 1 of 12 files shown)


wrongecho:
Test these changes at: https://customactions1079.pr-review.itflow.org  
 _(automatic message)_

johnnyq:
This is fantastic! Really nice integration on allowing custom hooks @wrongecho

PR #1090, opened 2024-10-08
Tidying

wrongecho:
- Move more things to new permissions system
- Deduplicate assets & docs post logic into model
- Swap out some "SELECT *" queries when only a couple of rows are actually needed


wrongecho:
Test these changes at: https://tidy1090.pr-review.itflow.org  
 _(automatic message)_

PR #1092, opened 2024-10-23
Update filter_header.php

ThaMunsta:
Change to filters for better experience and address an issue where by default its difficult to see new stripe expenses since they are added as "tomorrow" date and many filters exclude tomorrow. This will also take into account the users preference on start of week.

### DIFF
--- a/filter_header.php
+++ b/filter_header.php
@@ -71,12 +71,19 @@ if (!empty($_GET['sort'])) {
 // Date Handling
 if (empty($_GET['canned_date'])) {
     //Prevents lots of undefined variable errors.
-    // $dtf and $dtt will be set by the below else to 0000-00-00 / 9999-00-00
+    // $dtf and $dtt will be set by the below else to 0000-00-00 / end of the month
     $_GET['canned_date'] = 'custom';
 }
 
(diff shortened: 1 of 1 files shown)


wrongecho:
Test these changes at: https://defaultfilter1092.pr-review.itflow.org  
 _(automatic message)_

ThaMunsta:
A few edge cases still exist here. I believe if "today" is a Sunday (or Monday depending on user preferences) the this week and last week filters will be weird but otherwise I feel this has better filtering in most situations. Open to suggestions though.

johnnyq:
Hey @ThaMunsta I belive your timestamps might be in UTC instead of your Timezone, that could be why its off a bit

ThaMunsta:
Hey @johnnyq thanks, is that something I should update somewhere? The payment came through with the right date just the fees were put on the following day.
I still think this PR contains some nice QOL changes but I can just keep them in my local if they don't make sense to implement upstream. I could also make changes if you have any feedback. All good either way just love the software 😁

wrongecho:
I really like the ability to respect the Monday/Sunday week start preference if nothing else

PR #1121, opened 2024-12-14
API - New tickets

wrongecho:
- Allow creating tickets for no client / no contact
- Set URL view key
- Fix line break after ticket details that causes odd email formatting


wrongecho:
Test these changes at: https://apitckfix21121.pr-review.itflow.org  
 _(automatic message)_

Issue #1036, opened 2024-09-07
Feature Request: Enforce a Content Security Policy (CSP)


wrongecho:
Content Security Policies tell the browser to restrict where resources, like Javascript, are loaded from. They are a good defense against cross-site scripting (XSS) attacks.

We should try to move towards a CSP of at least `default-src 'self'`. Ideally adding `object-src 'none'` but these would require that we don't use Javascript in-line, but load it properly via scripts with hashing/nonces.

https://content-security-policy.com/faq/
https://content-security-policy.com/strict-dynamic/
https://content-security-policy.com/nonce/

Until we can add this as part of the header includes, we'll have to track individual pages that can and can't have CSP applied:

**General (main app)**

**Client Portal**
- Added to a few pages: index, invoices, login, profile, tickets

**Guest Views**

**Misc**

wrongecho:
Bumping this from Project board to an issue. Would appreciate any help with this.
Have started on some specific client portal pages but this seems to break things like TinyMCE.

It'd be fantastic (but probably a lot of work) to get this implemented in the main ITFlow app.

wrongecho:
Moved to https://tasks.dev.itflow.org/task_details.php?task_id=100

PR #1128, opened 2024-12-27
Stop using Source Sans Pro intermittently for some pages

wrongecho:
Starting to dig into merging the login flows and I noticed that we're still calling Source Sans Pro direct from Google. In [global.css](https://github.com/itflow-org/itflow/blob/a2f0f392be0f27ef5bb857a68efee6558102d4ed/global.css#L7) we define Sans Serif as the global font, I see no reason why we should have a slightly different font just for the login page. 
It looks like it was originally added sometime around when the login page was styled after being built and just never updated ([here](https://github.com/itflow-org/itflow/blob/e5036253edf7bc15e170f43d739314b32a68f599/login.php#L93)?).

By removing this we can also remove all references to fonts.googleapis.com fonts.gstatic.com as part of our CSP effort (#1036).

image

### DIFF
--- a/login.php
+++ b/login.php
@@ -1,7 +1,7 @@
 <?php
 
 // Enforce a Content Security Policy for security against cross-site scripting
-header("Content-Security-Policy: default-src 'self' fonts.googleapis.com fonts.gstatic.com");
+header("Content-Security-Policy: default-src 'self'");
 
 if (!file_exists('config.php')) {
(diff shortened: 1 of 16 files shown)


wrongecho:
Test these changes at: https://fonttidy1128.pr-review.itflow.org  
 _(automatic message)_

PR #1129, opened 2024-12-28
Tidying/General quality of life updates

wrongecho:
General quality of life updates:

- Ticket contact edit: 'No one' should be logged when ticket isn't associated with any contact, rather than a blank space; also fix email content
- Ticket portal/guest: Only show 'Assigned to' if someone is actually assigned
- Ticket tasks: Don't allow adding/modifying tasks for tickets in a resolved state
- Portal contacts: Properly align word 'Contacts' in a similar way all other page headings are aligned
- Client PDF export/handover: Now opens in a new self-closing window

### DIFF
--- a/client_download_pdf_modal.php
+++ b/client_download_pdf_modal.php
@@ -7,13 +7,13 @@
                     <span>&times;</span>
                 </button>
             </div>
-            <form action="post.php" method="post" autocomplete="off">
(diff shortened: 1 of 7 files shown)


wrongecho:
Test these changes at: https://tidying1129.pr-review.itflow.org  
 _(automatic message)_

PR #1148, opened 2025-01-12
Initial add Stripe Auto-payment with saved card

wrongecho:
- Ability for client to save/unsave card using the portal (Stripe customer is automatically created)
- Ability to automatically charge card via cron during recurring invoice / recurring invoice payment
- Ability to one-off charge the saved card for invoices


johnnyq:
Oh this is huge thanks for this @wrongecho

PR #1158, opened 2025-01-29
tickets views

ssteeltm:
Kanban view
Compact view


tickets.php
```diff
@@ -36,6 +36,18 @@
     }
 }
 
+if (isset($_GET['category'])) {
```
wrongecho, line 39: `Warning: Undefined variable $category_snippet in tickets.php on line 87`

This needs a blank assignment first incase category isn't in the GET like: `$category_snippet = '';`

ssteeltm: oops lets fix this

tickets.php
```diff
@@ -296,298 +350,29 @@ class="collapse
...
 
-                                if ($config_module_enable_accounting) {
-                                    require "modals/ticket_edit_billable_modal.php";
-                                }
-                            }
-                        }
+<?php
 
-                        ?>
+if ($_GET["view"]) {
```
wrongecho, line 358: `Warning: Undefined array key "view" in tickets.php on line 358`
This too.

Also, for the most part we do PHP logic at the very top of the page, unless it makes good sense to do otherwise

tickets_kanban.php
```diff
@@ -0,0 +1,310 @@
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/dragula/3.7.3/dragula.min.css">
```
wrongecho, line 1: Can we include the CSS locally rather than pulling it from a remote source please?

post/user/ticket_kanban.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+/*
+ * ITFlow - GET/POST request handler for client tickets
+ */
+
+defined('FROM_POST_HANDLER') || die("Direct file access is not allowed");
+
+if (isset($_POST['update_kanban_status_position'])) {
```
wrongecho, line 9: Asynchronous requests like update_kanban_status_position and update_kanban_ticket should be in ajax.php

tickets_kanban.php
ssteeltm: Sure, I can do this, also wih the JS file from dragula

post/user/ticket_kanban.php
ssteeltm: Believe me, I Thought about going there, but didnt know if it was old thing or not. I can change this.

tickets_kanban.php
```diff
@@ -0,0 +1,310 @@
...
+    }
+</style>
+
+
+<?php
+$kanban = [];
+
+
+// Fetch all statuses
+$status_sql = mysqli_query($mysqli, "SELECT * FROM ticket_statuses");
```
wrongecho, line 46: This should probably have a `where ticket_status_active = 1` to avoid showing deactivated statuses

tickets_kanban.php
johnnyq: these files should also be placed into the plugins/dragula folder

post/user/ticket_kanban.php
```diff
@@ -0,0 +1,50 @@
...
+        $ticket_id = intval($position['ticket_id']);
+        $kanban = intval($position['ticket_kanban']); // ticket kanban position
+        $status = intval($position['ticket_status']); // ticket statuses
+
+        // Continue if status is null
+        if ($status === null) {
+            continue;
+        }
+
+        mysqli_query($mysqli, "UPDATE tickets SET ticket_kanban = $kanban, ticket_status = $status WHERE ticket_id = $ticket_id");
```
wrongecho, line 43: Resolved tickets should be setting ticket_resolved_at, and un-resolving a ticket should unset it.

If $config_ticket_client_general_notifications is enabled, we should be sending the contact & any watchers an email to say it's resolved.

wrongecho:
Hey!

This is some really nice work here, well done!! I've left a couple comments for you to look over.

Generally speaking, I think we should exclude 'Closed' being a state on the board. I'm fine with resolved, but our stance with a ticket being closed is that you don't get to re-open it afterwards. Currently, the kanban bypasses the resolved -> closed flow and also lets you re-open tickets.
It probably also makes sense to hide the Kanban option when you are just viewing closed tickets (tickets.php?status=Closed) as you can't do anything with them.

Also, I'm thinking we should remove the ability to move columns around and rename `ticket_status_kanban` to something like `ticket_status_order` so it's set in the admin settings (like we do in `custom_links`). That way the admin determines the flow (New > Open > Hold > Custom > Resolved) rather than any random tech. We can use the order elsewhere too.

Thoughts? :)

post/user/ticket_kanban.php
ssteeltm: I see, I guess its best for now, to dont show Closed and Resolved Columns on Kanban. What do you think ?

ssteeltm:
> Hey!
> 
> This is some really nice work here, well done!! I've left a couple comments for you to look over.
> 
> Generally speaking, I think we should exclude 'Closed' being a state on the board. I'm fine with resolved, but our stance with a ticket being closed is that you don't get to re-open it afterwards. Currently, the kanban bypasses the resolved -> closed flow and also lets you re-open tickets. It probably also makes sense to hide the Kanban option when you are just viewing closed tickets (tickets.php?status=Closed) as you can't do anything with them.
> 
> Also, I'm thinking we should remove the ability to move columns around and rename `ticket_status_kanban` to something like `ticket_status_order` so it's set in the admin settings (like we do in `custom_links`). That way the admin determines the flow (New > Open > Hold > Custom > Resolved) rather than any random tech. We can use the order elsewhere too.
> 
> Thoughts? :)

I agree with you, but think we should hide also the resolved.
About the moving columns, liked ur ideia, but how about doing permissions on this? so only admin can move this columns

ssteeltm:
"t probably also makes sense to hide the Kanban option when you are just viewing closed tickets (tickets.php?status=Closed) as you can't do anything with them."

About this, makes all sense, will update to behave like that

tickets.php
ssteeltm: Lets keep like this for now. Later we can think a better way to do this?

post/user/ticket_kanban.php
ssteeltm: What do you sugest to have the notifications? As I see theres no function to do so, its all coded in ticket posts. So we would have to duplicate codes

wrongecho: Nice work setting the resolved at!

There's no function currently for the resolved email, so yeah we can just duplicate the code for now. It might be worth writing a function in the future.
My only other fear having resolved on the board and sending emails is that people will generate accidental messages stating tickets are resolved by mistakenly dropping a ticket in resolve and then moving it back out again.. @johnnyq - thoughts?

db.sql
```diff
@@ -2019,6 +2019,7 @@ CREATE TABLE `ticket_statuses` (
   `ticket_status_name` varchar(200) NOT NULL,
   `ticket_status_color` varchar(200) NOT NULL,
   `ticket_status_active` tinyint(1) NOT NULL DEFAULT 1,
+  `ticket_status_order` int(11),
```
wrongecho, line 2022: @johnnyq - Do we want the default for `ticket_status_order` to be 0 or NULL? Whatever we decide, it should be explicit.

db.sql
```diff
@@ -2113,6 +2114,7 @@ CREATE TABLE `tickets` (
   `ticket_asset_id` int(11) NOT NULL DEFAULT 0,
   `ticket_invoice_id` int(11) NOT NULL DEFAULT 0,
   `ticket_project_id` int(11) NOT NULL DEFAULT 0,
+  `ticket_kanban` int(11),
```
wrongecho, line 2117: @johnnyq - Do we want the default for `ticket_kanban` to be 0 or NULL? Whatever we decide, it should be explicit.

database_updates.php
```diff
@@ -2469,10 +2469,20 @@ function processFile($file_path, $file_name, $mysqli) {
         mysqli_query($mysqli, "UPDATE `settings` SET `config_current_database_version` = '1.8.0'");
     }
 
-    // if (CURRENT_DATABASE_VERSION == '1.8.0') {
-    //     // Insert queries here required to update to DB version 1.8.1
+    if (CURRENT_DATABASE_VERSION == '1.8.0') {
```
wrongecho, line 2472: also here to be updated with an explicit default declaration

db.sql
johnnyq: The default should be 0

tickets_kanban.php
```diff
@@ -0,0 +1,312 @@
...
+    <?php
+    }
+    ?>
+</div>
+
+
+    
+<script src="/plugins/dragula/dragula.min.js"></script>
+
+<script>
```
wrongecho, line 197: slight nickpick, but can we move this into `js/` and call it?

wrongecho:
Hey ssteeltm,

Nice changes, thank you! I'm really excited to get this reeled in.

We still need to make sure all output is sanitized and unfortunately SonarCloud doesn't catch everything. You can use htmlspecialchars (or our own nullable_htmlentities) to prevent [XSS injection](https://owasp.org/www-community/attacks/xss/) from asset names like this:
Screenshot 2025-02-03 202247

--

Is there a reason the Kanban doesn't show the ticket number and includes a blank circle at the top? I thought it was for client tags but apparently not?

Screenshot 2025-02-03 203325

--

> About the moving columns, liked ur ideia, but how about doing permissions on this? so only admin can move this columns

Potentially? We'll likely still want to add the ability to manage ticket_status_order via the admin settings and use it as an order by elsewhere.. Let's see what @johnny says.

db.sql
johnnyq: What do we need the field ticket_kanban for? I don't think this is actually needed

ssteeltm:
> Hey ssteeltm,
> 
> Nice changes, thank you! I'm really excited to get this reeled in.
> 
> We still need to make sure all output is sanitized and unfortunately SonarCloud doesn't catch everything. You can use htmlspecialchars (or our own nullable_htmlentities) to prevent [XSS injection](https://owasp.org/www-community/attacks/xss/) from asset names like this: Screenshot 2025-02-03 202247
> 
> --
> 
> Is there a reason the Kanban doesn't show the ticket number and includes a blank circle at the top? I thought it was for client tags but apparently not?
> 
> Screenshot 2025-02-03 203325
> 
> --
> 
> > About the moving columns, liked ur ideia, but how about doing permissions on this? so only admin can move this columns
> 
> Potentially? We'll likely still want to add the ability to manage ticket_status_order via the admin settings and use it as an order by elsewhere.. Let's see what @johnny says.

Hey! glad ur enjoying this like me! 
this blank if for category... here at least on kanban I didnt printed ticket id at all on screen, I want user can double click it so it opens

tickets_kanban.php
ssteeltm: Jhonny said somewhere that it was to go there on plugins... please tell me where to go and I can change

wrongecho: This bit is fine now that you've moved it to the plugins folder.
```<script src="/plugins/dragula/dragula.min.js"></script>```

--

Your _custom JS_ (below) needs moving to its own .js file and then including like above.

```
<script>
    $(document).ready(function() {
        // Initialize Dragula for the Kanban board
        var boardDrake = dragula([
            document.querySelector('#kanban-board')
        ], {
            moves: function(el, container, handle) {
                return handle.classList.contains('panel-title');
            }
        });
[...]
```

And ideally same for your custom css further up the page.

db.sql
johnnyq: @ssteeltm Can we sort it by something thats already there under tickets like sort by priority first then ticket_number or ticket_updated_at second? I don't know how I feel about adding an additional field just for ticket sorting in the kanban

ssteeltm: here I needed to put this, because like in Trello, users want in that collumn organize its ordering... I couldnt find anything to reuse ang get the same result...

johnnyq: I see, but lets remove that functionality of moving tickets within the existing Kanban status column. They should not be movable and sort by priority first then ticket number. But leave the functionality for dragging and dropping between the status columns.

post/user/ticket_kanban.php
johnnyq: Theres always an accident waiting to happen. hahaha I think its fine to move them to resolved and then back again. Ive done this in ticket details where I accidentally resolved and then reopened immediately.

db.sql
ssteeltm: Each of us wants slightly different behavior, so let’s make this configurable.

ssteeltm:
Guys, about Kanban Configs... do it under the Defaults item? or create new item Called Kanban ?
image

tickets_kanban.php
wrongecho: Ohhh, CSS _used_ to have its own folder but we moved it all under vendor/plugins. It can go in with `plugins/dragula/`, we don't need a dedicated folder just for it - just somewhere to keep it organised.

wrongecho:
Probably worth mentioning that our general principle is keeping things simple with good/opinionated defaults. In general we try and avoid making everything confusing and ultra-customisable. Therefore, I'd say we either include the ability to order tickets & move columns, or we don't. Not everything needs to be customisable.

I mentioned somewhere on the forum it'd be nice to add the default ticket view in settings. Defaults will be the best place for that but it might be best we add that separately after this is merged in the interests of getting this pulled in and tested as part of 25.02.

johnnyq:
> Probably worth mentioning that our general principle is keeping things simple with good/opinionated defaults. In general we try and avoid making everything confusing and ultra-customisable. Therefore, I'd say we either include the ability to order tickets & move columns, or we don't. Not everything needs to be customisable.
> 
> I mentioned somewhere on the forum it'd be nice to add the default ticket view in settings. Defaults will be the best place for that but it might be best we add that separately after this is merged in the interests of getting this pulled in and tested as part of 25.02.

I do like the default ticket view in the the ticket setting section. I think we should eventually move away from the defaults settings tab and move the entities in their respectable places such as invoice, ticket, quote etc settings. Its more logical.

db.sql
johnnyq: @ssteeltm I mean in that case lets just call it ticket_order, that way we can choose to use it in other views or places if we decide to do so later on.

tickets_kanban.php
ssteeltm: @wrongecho @johnnyq This css is related to the page tickets_kanban.php and not dragula itself. Still place it under dragula folder?

johnnyq:
Its looking good @ssteeltm im gonna reel it on in and we can make chnages if necessary during the Feb dev phase.

PR #1159, opened 2025-01-30
Beginning of the translation with the lang() function

Fr3ak88:
Beginning of the translation with the lang() function.

Three new files added.
lang.php represents the translation function.
en.php variables in English language
de.php variables in German language


wrongecho:
Hey & Welcome!

Is there a reason you're not using gettext for this as suggested in the forum?
It seems there are some advantages over a standard array, the main being that if we don't define a translation for _('something in english'), it just prints 'something in english' rather than a blank space.

See:
- https://stackoverflow.com/questions/2319060/using-php-gettext-extension-vs-php-arrays-in-multilingual-websites
- https://old.reddit.com/r/PHPhelp/comments/1688ict/handling_php_website_translation/jyu5vk5/

johnnyq:
Thank you for your efforts however we will be taking a different approach here in the future.

PR #1176, opened 2025-03-01
Update folder_create_modal.php

Whit3XLightning:
$folder['folder_name']; did not exist in the scope of where folder_create_modal.php was being required in the client_files.php and client_documents.php leading to null array pointer exception. While dirty the new way will reliably retrieve the name of the current folder the user is browsing and correctly name the modal.

image

image


johnnyq:
Nice that will work

johnnyq (approved):

wrongecho:
@johnnyq This should have been merged into develop, not master.

PR #1183, opened 2025-03-07
Setting added to pause ticket timer

whatinthemeh:
Just added a DB field and setting page option to have the ticket timer be paused by default instead of auto starting


wrongecho:
This is missing database update logic to add the field you're using.

PR #1190, opened 2025-03-20
Cron certificate notifs 

wrongecho:
Cron - Only notify on certificates expiring in 45 days if they are valid for longer than 90 days (i.e. not LE)

https://tasks.dev.itflow.org/task_details.php?task_id=55


johnnyq:
Nice I think the Let Encrypt Exception is a great idea, pulling

PR #1194, opened 2025-03-28
fix: missing kanban ticket settings

ssteeltm:
only today I've realized that I forgot to send this file, that will make settings of ticket view and kanban


wrongecho (approved):

johnnyq:
Nice pulling in

PR #1207, opened 2025-04-08
Fix Dashboard following DB Update 1.9.7

TamirSlo:
The `dashboard.php` file was still using old table names prior to DB version 1.9.7 - `recurring` instead of `recurring_invoices`. This meant the financial dashboard was not loading up correctly and an unhandled exception was thrown.

This commit fixes this and ensures the financial dashboard is displayed correctly.


johnnyq:
nice catch @TamirSlo Thank you

PR #1210, opened 2025-04-12
Update functions.php for Caddy

lilrags16:
This change gets the left most IP address from the X-Forward-For header. This is required to get the right IP easily from Caddy. Tested and working.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -79,7 +79,7 @@ function getUserAgent() {
 function getIP() {
     if (defined("CONST_GET_IP_METHOD")) {
         if (CONST_GET_IP_METHOD == "HTTP_X_FORWARDED_FOR") {
-            $ip = getenv('HTTP_X_FORWARDED_FOR');
+            $ip = explode(',', getenv('HTTP_X_FORWARDED_FOR'))[0];
         } else {
             $ip = $_SERVER["HTTP_CF_CONNECTING_IP"] ?? $_SERVER['REMOTE_ADDR'];
         }


lilrags16:
@johnnyq what is the process for getting PRs included upstream?

git-kup:
> @johnnyq what is the process for getting PRs included upstream?

https://docs.itflow.org/contribute

johnnyq:
how will this affect users not using Caddy?

lilrags16:
If someone is using the ```CONST_GET_IP_METHOD == "HTTP_X_FORWARDED_FOR"``` value in there config, it will always set the IP that ITFlow logs as the leftmost IP presented in a list of comma separated IPs.

Case 1 - Not using config value, not using caddy- No changes to use
Case 2 - Using config value, but reverse proxy only passes on IP: No effective change as the left most of a list of one is still the same
Case 3 - Caddy w/ Value set: The actually client IP is passed through correctly.

wrongecho:
@lilrags16 - Thanks for raising this, sorry it's taken so long to get some traction.
There's a merge conflict now but I've incluided your changes in #1222 amongst other things. We should always be grabbing the left-most IP if we're told there's a proxy in place. 
@johnnyq I've raised it as a PR into develop rather than a straight commit just to make you aware of it - it's worth testing across a variety of deployments (including the develop branch when its merged). It works fine on my local box and on my dev Hestia which has HTTP_X_FORWARDED_FOR defined.

PR #1211, opened 2025-04-15
1.Added Kenyan currency 2. Added docker implementation 

r-o-k-u:
1.Added Kenyan currency 2. Added docker implementation


wrongecho (changes requested):
Hello, thanks for the PR.

It doesn't need the docker file.

@johnnyq We should check KSH is actually supported by whatever function displays the currency symbols

PR #1222, opened 2025-06-10
Update how functions.php gets the remote IP address for logging

wrongecho:
- Builds on PR #1210 to always get the leftmost IP address
- Cloudflare (HTTP_CF_CONNECTING_IP) must now be explicitly defined, otherwise people could add the HTTP_CF_CONNECTING_IP header to a non-Cloudflare host and spoof IPs
- Tidy up the if/else logic a little


johnnyq (approved):
yeah lets merge this in, it makes sense to me

PR #1227, opened 2025-07-03
Update tickets_list.php

tysoe27:
Checks to see if status is set in the GET paramaters. Without this, a warning is shown (if PHP errors are enabled):

**Warning:** Undefined array key "status" in **tickets_list.php** on line **12**

### DIFF
--- a/tickets_list.php
+++ b/tickets_list.php
@@ -9,11 +9,11 @@
                         <tr>
                             
                             <td>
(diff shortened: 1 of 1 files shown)


johnnyq:
nice catch before I can merge a few suggestions you have a double } at the end and should use && instead of another of another if condition

PR #1230, opened 2025-07-16
setup.php fix

PowerChaos:
fixed mandatory currency and date setting, those are checked at the first page and you can only select them on the second page and that cause a 500 error on CWP

hard coded now the currency in the SQL file and the timezone in the PHP file so they wont be empty


wrongecho:
Hello,
Thanks for getting in touch.

I can't seem to reproduce an error 500 during setup. Could you provide details as to which page of setup triggers this specifically?

PR #1232, opened 2025-08-24
update.php

onlylegendleft:
This is a quick fix that should be edited to work with the comented code uncommented. I couldnt find a way to get create.php and update.php working without it. Its because both scripts are trying to use the same validation I think.

### DIFF
--- a/api/v1/tickets/update.php
+++ b/api/v1/tickets/update.php
@@ -14,24 +14,24 @@ $response = ['success' => false, 'message' => ''];
 //    die(json_encode(['error' => 'Cross-origin request denied']));
 //}
 
-// Validate and sanitize input
-$ticket_id = filter_input(INPUT_POST, 'ticket_id', FILTER_VALIDATE_INT, ['options' => ['min_range' => 1]]);
-if (!$ticket_id) {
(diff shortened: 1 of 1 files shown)


onlylegendleft:
I have an updated update.php ready to go with security patches. Just want to make sure this has been fixed to work the way we would like it to before moving on to security.

wrongecho:
Hi,

Thanks for your PR but I fail to understand why you need to edit the validate_api_key.php. This would break other API calls.

onlylegendleft:
> Hi,
> 
> Thanks for your PR but I fail to understand why you need to edit the validate_api_key.php. This would break other API calls.

It seems the validation does not work with update.php. Im still working on a fix at the moment. I have also had to edit multiple API files to fix the calls. I will post my progress here.

wrongecho:
You shouldn't need to be making all the changes, we have other update endpoints that work perfectly fine. You're doing something wrong.

onlylegendleft:
> You shouldn't need to be making all the changes, we have other update endpoints that work perfectly fine. You're doing something wrong.

I just tested without the change and it seems to be working without the validation edit.

onlylegendleft:
Security updates have been added.

onlylegendleft:
added logic for closed at and resolved at.

PR #1236, opened 2025-09-16
Update to use payment_methods table

cynicalgeek:
The Revenues modals were missed in commit c76da10747d9dbacbc49a8b190b6c07b96c1d7a8 that queries payment_methods instead of categories. These query changes fix the Payment Method dropdowns.


johnnyq:
nice catch! Ill pull this one in

PR #1237, opened 2025-09-16
Update invoice.php - removed extra unnecessary wording

git-kup:
it always bugged me that the alert message is so bloated. i almost never go to the mail queue after sending each invoice it just doesn't make any sense


johnnyq:
agreed and I think there are many places wording could use a little sprucing
Thank you pulling in

PR #1243, opened 2025-10-17
Exports and imports all fields

Thenezzar:
Exports and imports all fields in the database for credentials except the relations


agent/post/credential.php
```diff
@@ -401,38 +402,41 @@
         $duplicate_count = 0;
         while(($column = fgetcsv($file, 1000, ",")) !== false){
             $duplicate_detect = 0;
-            // Name
```
wrongecho: Please don't remove these comments, they're useful!!

wrongecho:
Hello, thanks for the PR and sorry for the delay.

Could you update the downloadable import CSV template to match the new fields you've added please - `download_credentials_csv_template`?

PR #1252, opened 2025-12-10
Develop Branch PR: Bug-fix: Nested Modals fix

RyanSawyer6003:
This PR fixed nested modal window behavior on close. Each modal now is in it's own container and parent modals are not closed on closing of child modals.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -401,3 +401,33 @@ $(document).ready(function() {
     // Data Tables
     new DataTable('.dataTables');
 });
+
+/*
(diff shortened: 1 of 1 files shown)


RyanSawyer6003:
Nevermind... I jumped the gun. This does work on closing nested modals, but not on saving new data. Closing PR will submit new when properly fixed.

johnnyq:
Your right you would almost have to refresh the parent model once the child modal submits so it can get any new data?

